PR #23, opened 2021-03-29
Add documentation validation workflow

LBGarber:
Adds `docs` workflow that runs on pull requests and commits. The workflow fails if CI-generated docs do not match with the repo's docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,36 @@
+name: Run Documentation Validation
+
+on: [push, pull_request]
+
+jobs:
+  run-docs-validation:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@v2
+
+      - name: update packages
+        run: sudo apt-get update -y
+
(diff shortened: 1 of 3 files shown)


.github/workflows/docs.yml
```diff
@@ -0,0 +1,39 @@
+name: Run Documentation Validation
+
+on:
+  workflow_dispatch:
+  push:
+  pull_request:
```
Charliekenney23, line 6: is this valid yaml?

.github/workflows/docs.yml
```diff
@@ -0,0 +1,39 @@
+name: Run Documentation Validation
+
+on:
+  workflow_dispatch:
```
Charliekenney23, line 4: not sure this is necessary

Charliekenney23 (changes requested):

.github/workflows/docs.yml
LBGarber: Should be valid, but it might be better to structure it as a list

Charliekenney23 (approved):

PR #30, opened 2021-04-07
Fix inconsistent type issue with no root password defined for instance

LBGarber:
This pull request fixes an issue that would error if an instance was created without a root password.


Charliekenney23 (changes requested):
Looks good to me. Can you add a changelog entry under minor changes of 0.0.4 before we merge this

Charliekenney23 (approved):

PR #31, opened 2021-04-07
Fix NodeBalancer diffing triggering on no changes

LBGarber:


plugins/module_utils/linode_helper.py
```diff
@@ -15,6 +15,9 @@ def dict_select_matching(d_1, d_2):
 
     return new_d1, new_d2
 
+def filter_null_values(input_dict):
+    """Returns a copy of the given dict with all keys containing nul values removed"""
```
Charliekenney23, line 19: typo

Charliekenney23 (changes requested):

Charliekenney23:
Changelog entry would be good for this, too.

plugins/module_utils/linode_helper.py
LBGarber: oops

PR #34, opened 2021-04-30
Add support for Linode Firewalls

LBGarber:


plugins/modules/firewall.py
```diff
@@ -0,0 +1,532 @@
...
+            result = self.client.networking.firewall_create(label, **spec_args)
+        except Exception as exception:
+            self.fail(msg='failed to create firewall: {0}'.format(exception))
+
+        return result
+
+    def __update_devices(self, spec_devices: list):
+        # Remove devices that are not present in config
+        remote_devices = self._firewall.devices
+
```
Charliekenney23, line 410: This could be a lot simpler:

Accumulate device IDs in sets, organized by type. This might looks like

```
{'linode': {123, 124}}
```

then you can iterate through the spec like:
```
if device_id not in remote_devices[device_type]:
  create_device(device_id, device_type)
  remote_devices[device_type].remove(device_id)

for device_type in remote_devices:
  for device_id in remote_devices[device_type]:
    # this device was unclaimed...
    delete(device_type, device_id)
```

Charliekenney23 (changes requested):

Charliekenney23 (approved):

PR #56, opened 2021-07-07
Use single leading underscore for private method names

LBGarber:
This change is necessary to prevent name mangling for module methods and to promote consistency with the rest of the collection.


Charliekenney23 (approved):
LGTM

PR #64, opened 2021-09-15
Default 'booted' to nil

LBGarber:
This default value was introduced in the documentation generation changes. If booted is set to `nil`, the collection defers boot handling to the API. This change will resolve the failing inventory integration test.


jriddle-linode (approved):
LGTM

Issue #73, opened 2021-10-04
[Help]: Cannot create multiple Domain records with the same name


onesign12:
### Description

Hi,

  - name: Create a A record
    linode.cloud.domain_record:
      domain: domain.com
      name: "{{ node_name }}"
      type: 'A'
      target: "{{ node_ip }}"
      port: 80
      state: present
    delegate_to: '127.0.0.1'

fatal: [.... -> 127.0.0.1]: FAILED! => {"changed": false, "msg": "failed to update domain record node2-10: type is a non-updatable field"}

LBGarber:
Hello!

Currently you will need to manually remove resources from your Linode account and re-run your playbook in order to recreate a resource with your desired value. This is to prevent the unintentional deletion and recreation of resources managed from Ansible. These mutable fields are currently enforced explicitly at an API level.

I definitely think the user experience could be improved here, maybe through some sort of `allow_recreation` field on each module. I'll add this to our internal tracker and investigate this further!

Let me know if you run into any other issues! :smiley:

onesign12:
Manually delete records in linode now A is ok but TXT failed 


TASK [Create a A record] *******************************************************
ok: [... -> 127.0.0.1]

TASK [Create a TXT record] *****************************************************
fatal: [... -> 127.0.0.1]: FAILED! => {"changed": false, "msg": "failed to update domain record node2-10: type is a non-updatable field"}

  - name: Create a A record
    linode.cloud.domain_record:
      domain: domain.com
      name: "{{ node_name }}"
      type: 'A'
      target: "{{ node_ip }}"
      port: 80
      state: present
    delegate_to: '127.0.0.1'


  - name: Create a TXT record
    linode.cloud.domain_record:
      domain: domain.com
      name: "{{ node_name }}"
      type: 'TXT'
      target: "v=spf1 a ip4:{{ drover_ip }} -all"
      port: 80
      state: present
    delegate_to: '127.0.0.1'


Second task trying update first one?

LBGarber:
Thanks for the update!

This seems to be an issue with the way we query records. Unfortunately, the lack of state management within Ansible makes it difficult to track records by their ID rather than `name` and `type`. I'll look into a solution for this and get back to you shortly!

iocouto:
Might be a nice idea to change the title of this issue to something a little more descriptive. 🙏

PR #74, opened 2021-10-04
Use `name`, `type`, and `target` to identify Domain Records 

LBGarber:
This change allows users to create Domain records with duplicate names and types. Additionally, users can now specify a `record_id` when updating an existing record.

#73


jriddle-linode (approved):
LGTM

PR #141, opened 2022-03-30
new: Add nodebalancer_node module

LBGarber:


l8gravely:
I'm going to try and test this out today or tomorrow.  Thanks for doing this work so quickly!  

One comment though, should there not also be a nodebalancer_config: module?  I think this is the one point of the linode_v4 api where things break down a bit, but I can understand where you have a nodebalancer and then the configs under it don't have a globablly unique ID, so you have to use the nodebalancer and config ids together when adding nodes.  

Annoying.  Too bad they couldn't just chain the IDs like <nodebalancerID>-<configID> or something, so you just have one ID to use.  Heck, I'm not even sure nodebalancer configs need to be more than 0, 1, 2 , .... , N in terms of IDs, based off the nodebalancer ID which is globally unique at Linode.  

So now a stupid question, what's the best way for me to test this branch out without over-writing the current version?  Or should I just install it on a seperate host to run my tests from?

LBGarber:
@l8gravely Glad you noticed this PR! 🙂 

> One comment though, should there not also be a nodebalancer_config: module? 

The reason I haven't yet been able to implement a `nodebalancer_config` module is there is no user-defined unique identifier that configs can be configured with. Because Ansible isn't stateful, there isn't a great way to implicitly store the ID of a config once it has been created. Tools like Terraform work around this issue by implicitly storing resource attributes, such as id, in a `tfstate` file.

I definitely think the inability to create configs independently of NodeBalancers in Ansible could hinder the user experience in some cases, so I'll keep looking into possible solutions.


> So now a stupid question, what's the best way for me to test this branch out without over-writing the current version? Or should I just install it on a seperate host to run my tests from?

Great question! 

I personally install Ansible inside of a `pyenv` environment for development purposes:

```sh
git clone -b new/nodebalancer_node https://github.com/linode/ansible_linode.git
cd ansible_linode

python3 -m venv ansible
source ansible/bin/activate

pip install -r requirements.txt -r requirements-dev.txt
make install

# Now you can run ansible-playbook, etc. from within the pyenv environment
```

jriddle-linode (approved):

PR #143, opened 2022-04-13
adding firewall_device module

jriddle-linode:
adding a module to allow altering firewall devices with an existing firewall not managed by ansible

```yaml
linode.cloud.firewall_device:
 firewall_id: 1234
 device_id: 12335465
 device_type: 'linode'
 state: present
```

```bash
make TEST_ARGS='-v firewall_device' test
```

### DIFF
--- a/plugins/modules/firewall_device.py
+++ b/plugins/modules/firewall_device.py
@@ -25,16 +25,22 @@ MODULE_SPEC = dict(
         description='The ID of the Firewall that contains this device.',
         ),
 
-    firewall_device_id=dict(
+    device_id=dict(
         type='int', required=True,
         description='The ID for this Firewall Device. This will be the ID of the Linode Entity.',
     ),
 
-    firewall_device_type=dict(
+    device_type=dict(
         type='str', required=True,
         description='The type of Linode Entity. Currently only supports linode.',
         choices=['linode'],
     ),
+
+    label=dict(
+        type='str',
+        required=False,
+        doc_hide=True,
+    )
 )
 
 specdoc_meta = dict(
@@ -65,7 +71,7 @@ class LinodeFirewallDevice(LinodeModuleBase):
         try:
             params = self.module.params
             firewall_id = params['firewall_id']
-            device_id = params['firewall_device_id']
+            device_id = params['device_id']
 
             firewall = linode_api4.Firewall(self.client, firewall_id)
             for device in firewall.devices:
@@ -76,13 +82,13 @@ class LinodeFirewallDevice(LinodeModuleBase):
         except Exception as exception:
             return self.fail(msg='failed to get device {0}: {1}'.format(device_id, exception))
 
-    def _create_node(self) -> linode_api4.FirewallDevice:
+    def _create_device(self) -> linode_api4.FirewallDevice:
         try:
             params = copy.deepcopy(self.module.params)
 
             firewall_id = params['firewall_id']
-            device_id = params['firewall_device_id']
-            device_type = params['firewall_device_type']
+            device_id = params['device_id']
+            device_type = params['device_type']
 
             firewall = linode_api4.Firewall(self.client, firewall_id)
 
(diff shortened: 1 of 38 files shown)


plugins/modules/firewall_device.py
```diff
@@ -0,0 +1,212 @@
...
+    firewall_device_id=dict(
+        type='int', required=True,
+        description='The ID for this Firewall Device. This will be the ID of the Linode Entity.',
+    ),
+
+    firewall_device_type=dict(
+        type='str', required=True,
+        description='The type of Linode Entity. Currently only supports linode.',
+        choices=['linode'],
+    ),
```
LBGarber, line 37: It might be a bit more intuitive to name these `device_id` and `device_type` 👍

plugins/modules/firewall_device.py
```diff
@@ -0,0 +1,212 @@
...
+            firewall = linode_api4.Firewall(self.client, firewall_id)
+            for device in firewall.devices:
+                if device.id == device_id:
+                    return device
+
+            return None
+        except Exception as exception:
+            return self.fail(msg='failed to get device {0}: {1}'.format(device_id, exception))
+
+    def _create_node(self) -> linode_api4.FirewallDevice:
```
LBGarber, line 79: Should this be `_create_device`?

LBGarber (changes requested):
Looks good assuming tests are passing! Just a few small changes.

You can run `make gendocs` to generate the `DOCUMENTATION` field and the corresponding docs RST.

jriddle-linode:
docs updated and mentioned requests changed, i went back and forth on device_id vs firewall_device_id like 10 times

l8gravely:
>>>>> "Jacob" == Jacob Riddle ***@***.***> writes:

Jacob> docs updated and mentioned requests changed, i went back and forth on device_id vs
Jacob> firewall_device_id like 10 times

fw_dev_id?  Maybe something shorter is ok?  Or fw_device_id?

Yeah, I'm just bikeshedding here...

John

LBGarber:
I'm still thinking on this myself. The Linode API does differentiate device/assignment IDs from entity IDs so it might be best to change the naming to prevent ambiguity. I wonder if simple `id` and `type` fields would be sufficient :thinking: 

Here's what we do in the Terraform provider: https://registry.terraform.io/providers/linode/linode/latest/docs/resources/firewall_device#argument-reference

API request: https://www.linode.com/docs/api/networking/#firewall-device-create__request-body-schema

plugins/modules/firewall_device.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def _get_device(self) -> Optional[linode_api4.FirewallDevice]:
+        try:
+            params = self.module.params
+            firewall_id = params['firewall_id']
+            entity_id = params['entity_id']
+
+            firewall = linode_api4.Firewall(self.client, firewall_id)
+            for device in firewall.devices:
+                if device.id == entity_id:
```
LBGarber, line 78:
```suggestion
            entity_id = params['entity_id']
            entity_type = params['entity_type']

            firewall = linode_api4.Firewall(self.client, firewall_id)
            for device in firewall.devices:
                if device.entity.id == entity_id and device.entity.type == entity_type:
```

tests/integration/targets/firewall_device/tasks/main.yaml
```diff
@@ -0,0 +1,60 @@
...
+        firewall_id: '{{ fw.firewall.id }}'
+        entity_id: '{{ inst.instance.id }}'
+        entity_type: 'linode'
+        state: present
+      register: fw_device
+
+    - name: Assert firewall device added
+      assert:
+        that:
+          - fw_device.changed
```
LBGarber, line 40: We might want to duplicate this test and assert `not fw_device.changed` to make sure we properly retrieve existing devices.

LBGarber (changes requested):
Looks good! One small thing and we should be good to go 🙂

LBGarber (approved):

PR #150, opened 2022-05-20
new: Overhaul docs generation system

LBGarber:


johnwcallahan (commented):
This is awesome, @LBGarber. The only bit of feedback I have is that, visually, the bullet point list of params is a little overwhelming for modules with a lot of arguments. Did you experiment with other options? Maybe a table? (This is a minor, subjective point -- not a blocker)

Screen Shot 2022-05-26 at 9 31 04 AM

johnwcallahan (approved):
@LBGarber that looks great 👍

PR #221, opened 2022-10-19
Feature/issue 216

marcfiu:
Draft PR for linode.cloud.firewall `state=update` feature. The focus of the `update` feature was for rules. The intent is for a workflow such that one can update specifically labeled rules without needing to know their position within the rules list. For example, a user may add arbitrary rules to a firewall (manually or using some other tool) and then one wants to use ansible to update specifically labeled rules.

### DIFF
--- a/plugins/modules/firewall.py
+++ b/plugins/modules/firewall.py
@@ -40,7 +40,7 @@ linode_firewall_rule_spec: dict = dict(
                description=[
                    'The label of this rule.'
                ]),
-    action=dict(type='str', required=True,
+    action=dict(type='str', choices=['ACCEPT', 'DROP'], required=True,
                 description=[
                     'Controls whether traffic is accepted or dropped by this rule.'
                 ]),
@@ -112,7 +112,7 @@ linode_firewall_spec: dict = dict(
                 ]),
     state=dict(type='str',
                description='The desired state of the target.',
-               choices=['present', 'absent'], required=True),
+               choices=['present', 'absent', 'update'], required=True),
 )
 
 
@@ -228,7 +228,7 @@ class LinodeFirewall(LinodeModuleBase):
         for rule in rules:
             item = copy.deepcopy(rule)
 
-            addresses = rule['addresses']
+            addresses = rule.get('addresses',[])
 
             if 'ipv6' in addresses:
                 item['addresses']['ipv6'] = [str(ipaddress.IPv6Network(v))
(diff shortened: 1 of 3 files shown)


tests/integration/targets/firewall_update/tasks/main.yaml
```diff
@@ -0,0 +1,613 @@
+tasks:
```
marcfiu, line 1: @LBGarber: I don't know how you run your test suite. I ran this manually separately with my own harness. For this reason, whatever I put here might not work out of the box with whatever test suite you use.

lgarber-akamai: We generally run the following command to execute the e2e test suite:
```
make TEST_ARGS="-v test_target_name" test
```

Fortunately this looks good for the most part, so it shouldn't be too much trouble to get it compatible.

plugins/modules/firewall.py
```diff
@@ -242,23 +242,57 @@ def _normalize_ips(rules: list) -> list:
 
         return result
 
-    def _rules_changed(self) -> bool:
-        """Returns whether the local and remote firewall rules match."""
+    def _amend_rules(self, remote_rules: list, local_rules: list) -> None:
```
lgarber-akamai, line 245: Small return type issue here
```suggestion
    def _amend_rules(self, remote_rules: list, local_rules: list) -> list:
```

plugins/modules/firewall.py
```diff
@@ -242,23 +242,57 @@ def _normalize_ips(rules: list) -> list:
...
+
+            # insert all missing fields in local_rule from remote_rule
+            local_rule = local_labeled_rules[remote_rule['label']]
+            for field in linode_firewall_rule_spec.keys():
+                if field not in local_rule:
+                    local_rule[field]=remote_rule[field]
+            result.append(local_rule)
+        return result
+
+    def _update_rules(self) -> list:
```
lgarber-akamai, line 265: It might be a bit cleaner to return an `Optional[dict]` here since we can return `None` when no rules need to be updated and the structure defined [here](https://www.linode.com/docs/api/networking/#firewall-rules-update__request-body-schema) if the rules need to be updated.

lgarber-akamai (changes requested):
Looks great so far! Just a few things I came across :slightly_smiling_face:

plugins/modules/firewall.py
```diff
@@ -242,23 +242,57 @@ def _normalize_ips(rules: list) -> list:
...
-
-            if field in remote_rules:
-                remote_rules[field] = self._normalize_ips(remote_rules[field])
-
-        return local_rules != remote_rules
+        for direction in ['inbound', 'outbound']:
+            local_rules[direction] = self._normalize_ips(local_rules[direction]) if direction in local_rules else []
+            remote_rules[direction] = self._normalize_ips(remote_rules[direction]) if direction in remote_rules else []
+
+        if self._state == "update":
```
lgarber-akamai, line 278: I wonder if there's a way we could break this out into another function or use guard clauses to make this a bit more readable. No big deal either way since it's primarily a preference and formatting thing :slightly_smiling_face:

plugins/modules/firewall.py
marcfiu: Do you mean for `_update_rules()` or the other method above? I'm guessing the former. Right now it returns an empty list if there are no changes, or a list of changes.  I do not believe it ever returns None.

marcfiu: Need clarification wrt the structured defined in the REST API. Previously the changed_rules() method this is based on returned a boolean and only upon True would the caller then actually update things. For this reason, maybe if this ever was supposed to return None to nullify a firewall it wasn't done that before either?

lgarber-akamai: Sorry for the late response!

> Do you mean for _update_rules() or the other method above? I'm guessing the former. Right now it returns an empty list if there are no changes, or a list of changes. I do not believe it ever returns None.

In this case I'm thinking more of a structural change. Since the [Firewall Rules Update](https://www.linode.com/docs/api/networking/#firewall-rules-update) endpoint accepts a dict describing the entire configuration of firewall rules (`inbound`, `inbound_policy`, `outbound`, and `outbound_policy`), I'm thinking it might be ideal for us to return `None` if there are no changes to be made and return the complete update request dict otherwise.

plugins/modules/firewall.py
marcfiu: I opted not to do that for now.

plugins/modules/firewall.py
marcfiu: _update_rules() always returns a dict, even an empty one.

_change_rules() will either return the new rules Dict or None if there are no changes.

Does this address your comment?

lgarber-akamai (approved):
Looks good to me! Waiting on a second review and we'll be good to merge

PR #229, opened 2022-11-08
Replace `str.removesuffix` usage in domain_record module

smkent:
`str.removesuffix` is new in Python 3.9:
https://docs.python.org/3/library/stdtypes.html#str.removesuffix

Python 3.7 and 3.8 are still in support:
https://devguide.python.org/versions/

On Python 3.7 or 3.8, the `domain_record` module fails with `failed to get domain record example.com: 'str' object has no attribute 'removesuffix'`

Python 3.8 is the default version of Python on Ubuntu 20.04 which is how this issue was discovered.

This change replaces the only usage of `str.removesuffix` in this repository with a conditional slice to restore compatibility with Python 3.7 and 3.8. This replacement provides the same functionality.

### DIFF
--- a/plugins/modules/domain_record.py
+++ b/plugins/modules/domain_record.py
@@ -140,11 +140,14 @@ class LinodeDomainRecord(LinodeModuleBase):
 
     def _find_record(self, domain: Domain, name: str, rtype: str, target: str) \
             -> Optional[DomainRecord]:
+        # We should strip the FQDN from the user-defined record name if defined.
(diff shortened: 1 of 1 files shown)


lgarber-akamai (approved):
Tested locally and all seems to work as intended. Thanks for the contribution!

Once this is merged, we can cut a patch release to unblock Python 3.7 and 3.8 users.

jriddle-linode (approved):

PR #231, opened 2022-11-08
vlans list module

jriddle-linode:
enable listing all vlans from `GET networking/vlans`


plugins/modules/vlans_list.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self) -> None:
+        self.module_arg_spec = spec
+        self.results: Dict[str, Any] = {
+            'vlans': []
+        }
+
+        super().__init__(module_arg_spec=self.module_arg_spec)
+
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for event list module"""
```
lgarber-akamai, line 83:
```suggestion
        """Entrypoint for VLAN list module"""
```

lgarber-akamai (changes requested):
Nice work! Could we rename this module to `vlan_list` to be inline with the other list modules?

lgarber-akamai (approved):
Tested locally and all looks good!

PR #254, opened 2023-01-09
Update Makefile to Utilize `LINODE_TOKEN` Environmental Variable

zliang-akamai:
Fix an issue that when `LINODE_API` environmental variable is set and `LINODE_API_API` isn't, `Make` would generate an `integration_config.yml` file with an empty API token.


lgarber-akamai (approved):
Looks good!

jriddle-linode (approved):

PR #256, opened 2023-01-11
new: Add improved timeout system for polling modules

lgarber-akamai:
## 📝 Description

This change implements a new context-based timeout system for modules that poll for resources statuses. This makes timeout periods significantly more predictable and reliable, and adds a standard way for handling timeouts in the future.

## ✔️ How to Test

```
make TEST_ARGS="-v instance_timeout" test
```

### DIFF
--- a/plugins/module_utils/linode_timeout.py
+++ b/plugins/module_utils/linode_timeout.py
@@ -1,26 +1,38 @@
+"""
+This module holds various utilities for managing module polling timeouts.
+"""
+
 import datetime
 
 
 class TimeoutContext:
+    """
+    TimeoutContext should be used by polling resources to track their provisioning time.
+    """
+
     def __init__(self, timeout_seconds=120):
         self._start_time = datetime.datetime.now()
         self._timeout_seconds = timeout_seconds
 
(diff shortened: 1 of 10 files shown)


jriddle-linode (approved):

plugins/module_utils/linode_timeout.py
```diff
@@ -0,0 +1,38 @@
...
+import datetime
+
+
+class TimeoutContext:
+    """
+    TimeoutContext should be used by polling resources to track their provisioning time.
+    """
+
+    def __init__(self, timeout_seconds=120):
+        self._start_time = datetime.datetime.now()
```
zliang-akamai, line 14: Maybe call `self.start` here? That would unify the way to start time.

zliang-akamai (approved):

plugins/module_utils/linode_timeout.py
lgarber-akamai: Good suggestion! I mainly avoided using `self.start` in `__init__` just so we can have an explicit definition of `_start_time` during initialization.

PR #259, opened 2023-01-12
issue-257: Remote rules may contain ipv4 or ipv6 addresses that

marcfiu:
are not being updated. This change focuses on increased fidelity to only update what's being updated. Added more unit tests and assertion to validate this assumption.

## 📝 Description

Firewall updates are intended to only make changes to specific firewall fields. A bug was found where if the firewall had an ipv6 address in one of its rules and the firewall update was just for ipv4, then all ipv6 addresses would be dropped/deleted.  Ditto vice versa when updating just ipv6 and not ipv4.

## ✔️ How to Test

The updated firewall_update/tasks/main.yml unit test validates correctness.

### DIFF
--- a/plugins/modules/firewall.py
+++ b/plugins/modules/firewall.py
@@ -262,6 +262,12 @@ class LinodeFirewall(LinodeModuleBase):
             for field in linode_firewall_rule_spec:
                 if field not in local_rule and field in remote_rule:
                     local_rule[field]=remote_rule[field]
+            for ip in ['ipv6','ipv4']:
+                if ip not in local_rule.get('addresses',{}):
+                    remote_addresses=remote_rule.get('addresses',{})
+                    local_addresses=local_rule.get('addresses',{})
+                    local_addresses[ip]=remote_addresses.get(ip,[])
+                    local_rule['addresses']=local_addresses
             result.append(local_rule)
         return result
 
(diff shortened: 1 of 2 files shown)


plugins/modules/firewall.py
```diff
@@ -262,6 +262,12 @@ def _amend_rules(remote_rules: list, local_rules: list) -> list:
             for field in linode_firewall_rule_spec:
                 if field not in local_rule and field in remote_rule:
                     local_rule[field]=remote_rule[field]
+            for ip in ['ipv6','ipv4']:
+                if ip not in local_rule.get('addresses',{}):
+                    remote_addresses=remote_rule.get('addresses',{})
+                    local_addresses=local_rule.get('addresses',{})
+                    local_addresses[ip]=remote_addresses.get(ip,[])
+                    local_rule['addresses']=local_addresses
```
lgarber-akamai, line 270: Small change to improve readability a bit
```suggestion
            for ip in ['ipv6', 'ipv4']:
                if ip in local_rule.get('addresses', {}):
                    continue
                    
                remote_addresses=remote_rule.get('addresses',{})
                local_addresses=local_rule.get('addresses',{})
                local_addresses[ip]=remote_addresses.get(ip,[])
                local_rule['addresses']=local_addresses
```

lgarber-akamai (changes requested):
PR looks great and all tests are passing on my end. Just one small stylistic change and addressing the linter feedback :slightly_smiling_face:

lgarber-akamai (approved):
Looks good!

marcfiu:
Btw., I've leveraged a jinja2 (macro + loops) to modularize the assertion and then from that produce the firewall_update/tasks/main.yml. Here's a gist link for that:
https://gist.github.com/marcfiu/5b66229f99550174f8456a532e1cd073

jriddle-linode (approved):
Looks good thank you for the contribution!

PR #272, opened 2023-01-23
Move `_get_resource_by_id` to the parent class

zliang-akamai:
Move the method to the paren class to reduce repeated code.

### DIFF
--- a/plugins/module_utils/linode_common.py
+++ b/plugins/module_utils/linode_common.py
@@ -6,19 +6,24 @@ import traceback
 from typing import Any, Type
 
 import polling
-
-from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import format_api_error
-from ansible_collections.linode.cloud.plugins.module_utils.linode_timeout import TimeoutContext
+from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import \
+    format_api_error
+from ansible_collections.linode.cloud.plugins.module_utils.linode_timeout import \
+    TimeoutContext
 
 try:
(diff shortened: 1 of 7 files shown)


jriddle-linode (approved):

plugins/module_utils/linode_common.py
```diff
@@ -141,6 +156,17 @@ def register_action(self, description: str) -> None:
         self.results['changed'] = True
         self.results['actions'].append(description)
 
+    def _get_resource_by_id(self, resource_type: Type[LinodeAPIType], resource_id: int):
+        try:
+            resource = resource_type(self.client, resource_id)
+            resource._api_get()
+            return resource
+        except Exception as exception:
+            resource_name = RESOURCE_NAMES.get(resource_type) or type(resource_type).__name__
```
lgarber-akamai, line 165: Nice! I wonder if it would be worth excluding the hard-coded `RESOURCE_NAMES` altogether so the function can be fully generic

lgarber-akamai (approved):
Nice work!

plugins/module_utils/linode_common.py
zliang-akamai: I am not sure about that. But the hardcoded dict does make the error message more clear by providing the names. In the future, we can put a `name` attribute in the API object of the Python API wrapper.

lgarber-akamai: Gotcha 👍 
I definitely think having a `display_name` attribute on resource classes in `linode_api4-python` would be the best approach for this going forward.

PR #279, opened 2023-01-27
fix: Fix `seconds_remaining` property method in TimeoutContext

lgarber-akamai:
## 📝 Description

This pull request corrects the `seconds_remaining` method to properly return the number of seconds remaining until the timeout expires rather than the amount of time since the timeout started.

This change should make the E2E test suite significantly more consistent.

## ✔️ How to Test

`make testall`


lgarber-akamai:
/acctest sha=8e90b9a5e5f5dcc7f861d8bf874866c12ac51d1a tests=lke_cluster_basic

PR #288, opened 2023-02-14
fix: Fix broken update logic in LKE module

lgarber-akamai:
## 📝 Description

This change resolves some broken update logic in the `lke_cluster` module.

## ✔️ How to Test

`make TEST_ARGS="-v lke_cluster_basic" test`

### DIFF
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -18,7 +18,7 @@ from ansible_collections.linode.cloud.plugins.module_utils.linode_common import 
 from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import global_authors, \
     global_requirements
 from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import \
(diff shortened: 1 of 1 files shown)


lgarber-akamai:
/acctest sha=e831258f50ebf78188ecd818b93fe21d97536204 tests=lke_cluster_basic

zliang-akamai (approved):

plugins/modules/lke_cluster.py
```diff
@@ -177,6 +177,15 @@
     'node_pools'
 }
 
+CREATE_FIELDS: Set[str] = {
```
zliang-akamai, line 180: Is this `CREATE_FIELDS` required to get it works? Do other modules need a similar thing or not?

lgarber-akamai: Good question! Basically, this is just a way of filtering out the meta-fields (e.g. `skip_polling`) from the actual resource fields (e.g. `node_pools`)

ezilber-akamai (approved):

PR #320, opened 2023-03-09
Added support for additional IP allocations in `linode_instance` module

ezilber-akamai:
## 📝 Description

Added support for additional IP allocations in `linode_instance` module. Additional IPv4 addresses can be allocated upon creation.

## ✔️ How to Test

`make TEST_ARGS="-v instance_basic" test`

### DIFF
--- a/plugins/modules/instance.py
+++ b/plugins/modules/instance.py
@@ -234,6 +234,12 @@ linode_instance_config_spec = dict(
     ),
 )
 
+spec_additional_ipv4 = dict(
+    public=SpecField(
+        type=FieldType.bool, description="fill this in", required=True
+    )
+)
+
 linode_instance_spec = dict(
     type=SpecField(
         type=FieldType.string,
@@ -374,7 +380,8 @@ linode_instance_spec = dict(
     ),
     additional_ipv4=SpecField(
         type=FieldType.list,
-        element_type=FieldType.string,
+        element_type=FieldType.dict,
+        suboptions=spec_additional_ipv4,
         description=["Additional ipv4 addresses to allocate."],
         editable=False,
     ),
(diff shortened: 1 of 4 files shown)


ezilber-akamai:
/acctest tests=instance_basic sha=a1d4924b2a0a9e998e9651a222302fcd01299e15

plugins/modules/instance.py
```diff
@@ -926,6 +940,15 @@ def _handle_present(self) -> None:
...
                 )
+
+            if self.module.params.get("additional_ipv4") is not None:
+                additional_ip_types = self.module.params.get("additional_ipv4")
+
+                for ip_type in additional_ip_types:
+                    if ip_type == "public":
+                        self._instance.ip_allocate(public=True)
+                    elif ip_type == "private":
+                        self._instance.ip_allocate(public=False)
```
lgarber-akamai, line 951: This could simplify down a bit:
```
self._instance.ip_allocate(public=ip_type=="public")
```

tests/integration/targets/instance_basic/tasks/main.yaml
```diff
@@ -34,6 +54,35 @@
...
+        image: linode/ubuntu20.04
+        private_ip: true
+        wait: false
+        state: present
+        additional_ipv4:
+          - "public"
+          - "public"
+      register: invalid_update_add_ips
+      failed_when:
+        - invalid_update_add_ips.changed == true
```
lgarber-akamai, line 71: Maybe we could check to see if an error is raised here?

plugins/modules/instance.py
```diff
@@ -372,6 +372,12 @@
             'have status "running".'
         ],
     ),
+    additional_ipv4=SpecField(
```
lgarber-akamai, line 375: We can use the `choices` field here to only allow `public` or `private`. 

I also wonder if this could be made a bit more future-proof by making this field a list of dicts. This might come in handy if we need configuration fields in the future 👍 

e.g. 
```
additional_ipv4:
  - public: true
  - public: false
```

Take a look at the [firewall module](https://github.com/linode/ansible_linode/blob/dev/plugins/modules/firewall.py#L134) for a good example of how the spec entry would look.

lgarber-akamai (changes requested):
Great work so far! Just a few suggestions

ezilber-akamai:
/acctest tests=instance_basic sha=b0b6ba2b058a40ecabe1d980b10c9898ce3bc065

plugins/modules/instance.py
zliang-akamai:
```
additional_ipv4:
  - public: true
  - public: false
```

Does something like this makes it have 2 additional IP addresses?

plugins/modules/instance.py
```diff
@@ -851,6 +864,23 @@ def _update_instance(self) -> None:
         if should_update:
             self._instance.save()
 
+        ipv4_length = len(self.module.params.get("additional_ipv4") or [])
```
zliang-akamai, line 867:
```suggestion
        ipv4_length = len(self.module.params.get("additional_ipv4", []))
```
Optional suggestion. It might be more readable this way.

ezilber-akamai: I tried doing it this way and it didn't work. I think this is because `additional_ipv4` is an optional parameter, so if it isn't set `len` is run on a `NoneType`, causing the test to fail.

zliang-akamai: Great, that makes sense!

plugins/modules/instance.py
```diff
@@ -851,6 +864,23 @@ def _update_instance(self) -> None:
...
+                    )
+                )
+        else:
+            if ipv4_length != len(getattr(self._instance, "ipv4")) - 1:
+                self.fail(
+                    "failed to update instance {0}:additional_ipv4 is a non-updatable field".format(
+                        self._instance.label
+                    )
+                )
+
```
zliang-akamai, line 883: Since they would throw the same error message, can you try to combine these 2 if-else branches?
You can try something like:
```python
min_ips = 2 if self.module.params.get("private_ip") else 1
if ipv4_length != len(getattr(self._instance, "ipv4")) - min_ips:
    ...
```

ezilber-akamai: Thanks! Took care of this.

zliang-akamai (approved):

plugins/modules/instance.py
```diff
@@ -234,6 +234,12 @@
     ),
 )
 
+spec_additional_ipv4 = dict(
+    public=SpecField(
+        type=FieldType.bool, description="fill this in", required=True
```
lgarber-akamai, line 239: Looks like the description still needs to be written here

lgarber-akamai (approved):
Nice work! Just one small thing but I'll go ahead and preemptively approve

plugins/modules/instance.py
ezilber-akamai: Good catch, totally forgot about this!

PR #323, opened 2023-03-28
Added stats module

ezilber-akamai:
## 📝 Description

Added `nodebalancer_statistics` ansible module.

## ✔️ How to Test

`make TEST_ARGS="-v nodebalancer_stats" test`

### DIFF
--- a/plugins/modules/nodebalancer_stats.py
+++ b/plugins/modules/nodebalancer_stats.py
@@ -1,0 +1,120 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""This module allows users to view Nodebalancer Stats."""
+
+from __future__ import absolute_import, division, print_function
+
+from typing import Any, List, Optional
+
+import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.nodebalancer_statistics as docs
+from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
+    LinodeModuleBase,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import (
+    global_authors,
+    global_requirements,
+)
+from ansible_specdoc.objects import (
+    FieldType,
+    SpecDocMeta,
+    SpecField,
+    SpecReturnValue,
+)
+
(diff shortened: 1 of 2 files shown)


plugins/modules/nodebalancer_stats.py
```diff
@@ -0,0 +1,126 @@
...
+        type=FieldType.dict,
+        suboptions=traffic_spec,
+        description=["Traffic statistics for this NodeBalancer."],
+    ),
+    title=SpecField(
+        type=FieldType.string,
+        description=[
+            "The title for the statistics generated in this response."
+        ],
+    ),
```
lgarber-akamai, line 67: Since these are all response fields, they can be excluded module spec 👍

lgarber-akamai (changes requested):
Great work! Just one small change

plugins/modules/nodebalancer_stats.py
ezilber-akamai: Does this mean that id is the only field that should be in the module spec?

plugins/modules/nodebalancer_stats.py
```diff
@@ -0,0 +1,93 @@
...
+
+        super().__init__(
+            module_arg_spec=self.module_arg_spec,
+            required_one_of=self.required_one_of,
+        )
+
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for NodeBalancer Statistics module"""
+
+        self.results["node_balancer_stats"] = self.client.get(
```
lgarber-akamai, line 80: Could this logic be adjusted to resolve the NodeBalancer from either the label or the ID?
See: https://github.com/linode/ansible_linode/blob/dev/plugins/modules/image_info.py#L97

lgarber-akamai (changes requested):

lgarber-akamai:
/acctest sha=3348123abe8cc77be4a0952bddf8a3a8e1929d34 tests=nodebalancer_stats

zliang-akamai (approved):
Looks good to me!

PR #327, opened 2023-03-31
Add support for volume cloning

ezilber-akamai:
## 📝 Description

Add support for volume cloning in volume module

## ✔️ How to Test

`make TEST_ARGS="-v volume_basic" test`

Ticket: TPT-1850

### DIFF
--- a/plugins/modules/volume.py
+++ b/plugins/modules/volume.py
@@ -23,6 +23,12 @@ from ansible_specdoc.objects import (
     SpecField,
     SpecReturnValue,
 )
+from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import (
+    request_retry,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_event_poller import (
+    EventPoller,
+)
 from linode_api4 import Volume
 
 linode_volume_spec = dict(
@@ -89,6 +95,11 @@ linode_volume_spec = dict(
         choices=["present", "absent"],
         required=True,
     ),
+    source_volume_id=SpecField(
+        type=FieldType.integer,
+        required=False,
+        description=["The volume id of the desired volume to clone."],
+    )
 )
 
 SPECDOC_META = SpecDocMeta(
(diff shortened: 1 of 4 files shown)


jriddle-linode (approved):

plugins/modules/volume.py
```diff
@@ -175,14 +186,99 @@ def poll_func() -> bool:
...
+        cloned_volume = Volume(self.client, vol.get("id"))
+        cloned_volume._api_get()  # Force lazy-loading
+
+        self.register_action(
+            "Cloned volume with source id {0}".format(source_id)
+        )
+        self._wait_for_volume_active(cloned_volume)
+
+        # Resize if necessary
+        if resize:
```
yec-akamai, line 235: The `resize` flag seems only be used here. Should we do 
`if params.get("size") != source_volume.size` directly instead of storing the boolean value?

plugins/modules/volume.py
```diff
@@ -175,14 +186,99 @@ def poll_func() -> bool:
                 "failed to wait for volume status: timeout period expired"
             )
 
-    def _wait_for_volume_active(self) -> None:
+    def _wait_for_volume_active(self, volume: Volume) -> None:
```
yec-akamai, line 189: How about we make the `volume` param optional to avoid some code change when calling this method?

ezilber-akamai: The calls to `_wait_for_volume_active(self)` in the `_clone_volume` function require a volume parameter to work properly and the implementation of the function had to change slightly so that it would not work without the volume parameter. I tried changing it to optional and the tests all pass but I think that if we tried to make a call to `_wait_for_volume_active` without a volume parameter, even if it is optional, there would be an error.

yec-akamai: It's true that when cloning the volume we need the `volume` parameter, however, in other cases, i.e. `Ensure volume is active before continuing`, the param is not required. If we can make it optional, we don't have to change the legacy code or pass the volume param when it can be got from self. It can offer more flexibility and simplicity on how the function can be called. We can modify implementation of this function to tell that it can be used in both case. My intuitive thinking is like:

```
def _wait_for_volume_active(self, volume: Volume = None) -> None:
        # if volume is not passed, get its value from self (maybe also add the scenarios when it could be passed or not)
        if volume is None: 
             volume = self._volume
        self._wait_for_volume_status(
            volume, {"active"}, self._timeout_ctx.seconds_remaining
```

Let me know how you think about it!

ezilber-akamai: I tried this and this implementation caused errors. I'm not positive, but I think it was because self._volume is only populated with the correct volume in the function that calls _wait_for_volume_active rather than in _wait_for_volume_active itself. So, even though volume is set to self._volume when no argument is passed, self._volume is None at the point where it is assigned. This makes me think that the volume must be passed in the legacy calls because that's the only point at which we can access it through self._volume. What are your thoughts on this?

plugins/modules/volume.py
```diff
@@ -175,11 +186,48 @@ def poll_func() -> bool:
...
+            )
         )
 
+        cloned_volume = Volume(self.client, vol.get("id"))
+        cloned_volume._api_get()  # Force lazy-loading
+
+        self.register_action(
+            "Cloned volume with source id {0}".format(source_id)
+        )
+        self._wait_for_volume_active(cloned_volume)
```
zliang-akamai, line 227: Can we let line 252 call `_wait_for_volume_active`, after the volume is assigned to `self._volume`, similar to what `_create_volume` handles it?

plugins/modules/volume.py
zliang-akamai: I saw `self._volume` is assigned in line 246 or 248. So, how`self._volume` can be `None` when `_wait_for_volume_active` is call in line 252 and 256?

lgarber-akamai: Personally I'm in favor of the volume needing to be explicitly passed in as an argument since it requires that a volume must already exist before the function is called, which can improve readability and reduce ambiguity a bit. I think the issue we're running into right now is because of the fact that it's not clear whether or not the volume has been stored in `self._volume` at the time of calling. 

I originally wrote this module very early on in this project and at the time thought it'd be best to store the current volume in `self._volume`, but I think passing it around as a variable makes the code quite a bit cleaner. 👍

ezilber-akamai: > I saw `self._volume` is assigned in line 246 or 248. So, how `self._volume` can be `None` when `_wait_for_volume_active` is call in line 252 and 256?

This was happening after I made some code changes to make `volume` and optional parameter for `_wait_for_volume_active`. The code currently in the PR does not have this issue. Sorry for the confusion!

plugins/modules/volume.py
ezilber-akamai: The way `_create_volume` is currently written, I cannot call `_wait_for_volume_active` there because the volume is created in a one-liner return statement. I can modify the code so that it spans multiple lines but I think it would make the function unnecessarily bloated. What do you think?

yec-akamai (approved):

plugins/modules/volume.py
ezilber-akamai: Took care of this!

plugins/modules/volume.py
```diff
@@ -175,11 +186,47 @@ def poll_func() -> bool:
...
+                data={"label": params.get("label")},
+            )
         )
 
+        cloned_volume = Volume(self.client, vol.get("id"))
+        cloned_volume._api_get()  # Force lazy-loading
+
+        self.register_action(
+            "Cloned volume with source id {0}".format(source_id)
+        )
```
zliang-akamai, line 226: Can we move this to somewhere after line 245? Also, I think it may be nicer if we include the created volume's label in the action message.

zliang-akamai:
make TEST_ARGS="-v volume_basic" test

zliang-akamai:
/acctest sha=0d2ca201ced5b66bcab293ffba4e77411f51c36c tests=volume_basic

zliang-akamai:
/acctest sha=de9d8cd4ecbf32192700d002e7b43677ea2b31a9 tests=volume_basic

zliang-akamai (approved):
Nice work!

zliang-akamai:
/acctest sha=de9d8cd4ecbf32192700d002e7b43677ea2b31a9 tests=volume_basic

ezilber-akamai:
/acctest sha=de9d8cd4ecbf32192700d002e7b43677ea2b31a9 tests=volume_basic

PR #352, opened 2023-05-04
Add backups_enabled option to instance creation fields

godismyjudge95:
## 📝 Description

Linode supports enabling backups upon creation of an instance:  https://www.linode.com/docs/api/linode-instances/#linode-create__request-body-schema

This PR adds this option to the instance fields


lgarber-akamai:
@godismyjudge95 Thanks for the contribution! Do you mind if I push up some accompanying tests and generated documentation.

lgarber-akamai (changes requested):
We can also implement update logic to enroll/cancel backups for a Linode.

This would require:
- Removing `default=False` so users can avoid accidentally disabling backups on existing instances
- Adding `backups_enabled` to the list of exceptions [here](https://github.com/linode/ansible_linode/blob/dev/plugins/modules/instance.py#L834)
- Add logic [here](https://github.com/linode/ansible_linode/blob/dev/plugins/modules/instance.py#L869) to:
  -  Check if `backups_enabled` is not None and `backups_enabled` != `self._instance.backups.enabled`
  - If the above condition is met, enable backups if `backups_enabled` is True else disable backups.

godismyjudge95:
@lgarber-akamai Feel free to make changes, I just noticed this was missing the other day and wanted to get the ball rolling.

zliang-akamai:
Thanks for bringing this up! We will continue working on it on a new branch.

godismyjudge95:
> Thanks for bringing this up! We will continue working on it on a new branch.

Thank you for continuing implementing it, I'll be on the lookout for when this goes live - could really use it in our configs :)

zliang-akamai:
@godismyjudge95 We previously released the newer version of  Linode Ansible collection, and I think you can now use `backups_enabled` option with the latest version of the collection. Let us know if there is any question or issue, ^_^

PR #355, opened 2023-05-09
new: Add `rebooted` field to the `instance` module

### LINKED ISSUE #336: [Feature]: Linode reboot
### Description

Option in _linode.cloud.instance_ to force reboot a linode.

This would be usefull in cases where you use Ansible to apply updates and reboot machines and they hang on booting process.
This then can be used as a _rescue_ action from a _block_.

### New/Affected Components

linode.cloud.instance

### Example Ansible Config

- name: Reboot a Linode instance.
  linode.cloud.instance:
    label: my-linode or id: <linode id>
    state: rebooted or action: reboot

lgarber-akamai:
## 📝 Description

This change adds a `rebooted` field to the `instance` module that allows users to automatically reboot running Linode instances.

Resolves #336 

## ✔️ How to Test

```
make TEST_ARGS="-v instance_reboot" test
```


zliang-akamai (approved):
Overall looks good!

PR #389, opened 2023-07-12
new: Add Python dependency validation and enforcement

lgarber-akamai:
## 📝 Description

This change adds automatic dependency enforcement treating `requirements.txt` as the single source of truth. This allows for a better user-experience by informing users when their Python dependency versions are not compatible with the current collection version.

Due to limitations outlined in `scripts/embed_deps.py`, requirements are automatically embedded into the collection as a small Python module that is imported from `linode_common.py`. This module is automatically generated and embedded when the `make embed-deps` target is run.

Additionally, this change introduces a new GitHub actions workflow to ensure the embedded dependency module matches the dependencies found in `requirements.txt`.

### DIFF
--- a/.github/workflows/embedded-deps.yml
+++ b/.github/workflows/embedded-deps.yml
@@ -1,0 +1,44 @@
+name: Run Embedded Requirements Validation
+
+on: pull_request
+
+jobs:
+  run-req-validation:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: .ansible/collections/ansible_collections/linode/cloud
+    steps:
+      - name: checkout repo
+        uses: actions/checkout@v3
+        with:
+          path: .ansible/collections/ansible_collections/linode/cloud
+
+      - name: update packages
+        run: sudo apt-get update -y
+
+      - name: install packages
+        run: sudo apt-get install -y make
+
+      - name: setup python 3
+        uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
+
+      - name: install dependencies
+        run: pip install -r requirements-dev.txt -r requirements.txt
+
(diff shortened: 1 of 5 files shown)


jriddle-linode (approved):

zliang-akamai:
[`pkg_resources` is now a deprecated API](https://setuptools.pypa.io/en/latest/pkg_resources.html), let's use `importlib` instead. By the way, `requirements.txt` format is [generally not the most favorable way](https://stackoverflow.com/a/59971236/22181775) to define the dependencies today, as the community is moving to `pyproject.toml`. The requirements file API moved to [`packaging`](https://github.com/pypa/packaging) but it's not in Python standard library. I am still trying to find a best approach for it.

lgarber-akamai:
> [`pkg_resources` is now a deprecated API](https://setuptools.pypa.io/en/latest/pkg_resources.html), let's use `importlib` instead. By the way, `requirements.txt` format is [generally not the most favorable way](https://stackoverflow.com/a/59971236/22181775) to define the dependencies today, as the community is moving to `pyproject.toml`. The requirements file API moved to [`packaging`](https://github.com/pypa/packaging) but it's not in Python standard library. I am still trying to find a best approach for it.

Good catch! This is a bit tricky to approach because of the limitations you mentioned above, as well as the fact that we'll need to maintain a SSOT for users to install dependencies from. 

One factor we should take into consideration is that this project isn't distributed as a Python package but rather an Ansible-Galaxy `.tar.gz` archive, so all dependency installation will need to be handled manually by the user.

lgarber-akamai:
Holding off on this PR until we discuss it as a team.

lgarber-akamai:
Rejecting in favor of a "vendor" directory

PR #401, opened 2023-08-09
new: Add `ip_share` module

yec-akamai:
## 📝 Description

Add the `ip_share` module and related docs. This module allow customer to configure IPs that can be shared with the target Linode. It makes calls to the [IP addresses share ](https://www.linode.com/docs/api/networking/#ip-addresses-share)endpoint. 

## ✔️ How to Test

make TEST_ARGS="-v ip_share" test


plugins/modules/ip_share.py
```diff
@@ -0,0 +1,111 @@
...
+                msg="failed to configure shared ips for linode {0}: {1}".format(
+                    self.module.params.get("linode_id"), exception
+                )
+            )
+
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for configuring shared IPs for a Linode."""
+        linode_id = self.module.params.get("linode_id")
+
+        self._share_ip_addresses()
```
lgarber-akamai, line 93: To stay consistent with other modules, could we check that the IPs have not been shared with the target Linode and call `self.register_action(...)` if so. This will automatically populate the `changed` field which will reflect in the output of an `ansible-playbook` run.

lgarber-akamai (changes requested):
Great work so far!

jriddle-linode (approved):

plugins/modules/ip_share.py
```diff
@@ -0,0 +1,159 @@
...
+                    linode_id, exception
+                )
+            )
+
+    # check if the IPs have been shared with the Linode instance
+    def _check_shared_ip_addresses(
+        self, ips: List[str], linode: Instance
+    ) -> bool:
+        linode_shared_ipv4 = [i.address for i in linode.ips.ipv4.shared]
+        linode_public_ipv6 = [i.range for i in linode.ips.ipv6.ranges]
```
lgarber-akamai, line 101: To ensure the returned IPv6 ranges are only shared, could ensure that `is_bgp` is `True` for each of these ipv6 ranges? You might need to make a few extra network requests to the [IPv6 Range View](https://www.linode.com/docs/api/networking/#ipv6-range-view) endpoint since `is_bgp` seems to be missing from some IPv6 response bodies.

lgarber-akamai (changes requested):

plugins/modules/ip_share.py
yec-akamai: Good catch! I added this check to ensure `is_bgp` is True for these ipv6 ranges.

plugins/modules/ip_share.py
```diff
@@ -0,0 +1,164 @@
...
+        linode_shared_ipv6 = []
+        for ipv6 in [i.range for i in linode.ips.ipv6.ranges]:
+            if ipv6.is_bgp:
+                linode_shared_ipv6.append(ipv6)
+
+        for i in ips:
+            if i not in linode_shared_ipv4 and i not in linode_shared_ipv6:
+                return False
+
+        return True
```
lgarber-akamai, line 112: This function can be simplified down a bit:
```suggestion
    # check if the IPs have been shared with the Linode instance
    def _check_shared_ip_addresses(
        self, ips: List[str], linode: Instance
    ) -> bool:
        current_ips = {i.address for i in linode.ips.ipv4.shared}

        # ensure that IPv6 ranges are only shared by checking if is_bgp is True
        for ipv6 in [i for i in linode.ips.ipv6.ranges]:
            # We need to make a manual GET request
            # because is_bgp is only available in the GET
            # response body.
            ipv6._api_get()

            if ipv6.is_bgp:
                current_ips.add(ipv6.range)

        return set(ips) == current_ips
```

tests/integration/targets/ip_share/tasks/main.yaml
```diff
@@ -0,0 +1,72 @@
...
+        state: present
+      register: ip_shared
+
+    - name: Configure a Linode with already shared IPs.
+      linode.cloud.ip_share:
+        api_version: v4beta
+        ips: ['{{ instance_create.instance.ipv4[0]}}']
+        linode_id: '{{ instance_create_shared.instance.id }}'
+        state: present
+      register: ip_already_shared
```
lgarber-akamai, line 42: Could this also test that an IPv6 range will be shared as expected? It's a bit tricky since we don't have an IPv6 address module, but we can use the `api_request` module like this:

```suggestion

    - name: Create an IPv6 range
      linode.cloud.api_request:
        method: POST
        path: networking/ipv6/ranges
        body_json: >
          {
            "linode_id": {{ instance_create.instance.id }},
            "prefix_length": 64
          }
      register: create_ipv6

    - set_fact:
        ipv6_range: '{{ (create_ipv6.body.range | split("/"))[0] }}'

    # IPv6 ranges must be shared with their assigned Linode before they can be shared to other Linodes
    - name: Share the IPv6 range to the assigned Linode.
      linode.cloud.ip_share:
        api_version: v4beta
        ips: ['{{ ipv6_range }}']
        linode_id: '{{ instance_create.instance.id }}'
        state: present
      register: ipv6_shared

    - name: Configure a Linode shared IPs.
      linode.cloud.ip_share:
        api_version: v4beta
        ips: ['{{ instance_create.instance.ipv4[0]}}', '{{ ipv6_range }}']
        linode_id: '{{ instance_create_shared.instance.id }}'
        state: present
      register: ip_shared

    - name: Configure a Linode with already shared IPs.
      linode.cloud.ip_share:
        api_version: v4beta
        ips: ['{{ instance_create.instance.ipv4[0]}}', '{{ ipv6_range }}']
        linode_id: '{{ instance_create_shared.instance.id }}'
        state: present
      register: ip_already_shared
```

lgarber-akamai (changes requested):
Thanks for adding the `is_bgp` check! Just a few more things and this PR should be ready to go :tada:

plugins/modules/ip_share.py
yec-akamai: Thanks for the code suggestion!

lgarber-akamai (approved):
Works great, nice work!

PR #402, opened 2023-08-15
new: `ip_assign` module

jriddle-linode:
## 📝 Description

adding [ip_assign](https://www.linode.com/docs/api/networking/#ip-addresses-assign) module

## ✔️ How to Test

**What are the steps to reproduce the issue or verify the changes?**

`make TEST_ARGS="-v ip_assign" test`

### DIFF
--- a/plugins/module_utils/doc_fragments/ip_assign.py
+++ b/plugins/module_utils/doc_fragments/ip_assign.py
@@ -1,0 +1,11 @@
+"""Documentation fragments for the ip_assign module"""
+specdoc_examples = ['''
+- name: Assign IP to Linode
+  linode.cloud.ip_assign:
+  linode.cloud.ip_assign:
+    region: us-east
+    assignments:
+     - address: 0.0.0.0
+       linode_id: 123''']
+
+result_ip_assign_samples = ['''[{}]''']
--- a/plugins/module_utils/doc_fragments/ip_assign.py
+++ b/plugins/module_utils/doc_fragments/ip_assign.py
@@ -2,7 +2,6 @@
 specdoc_examples = ['''
 - name: Assign IP to Linode
   linode.cloud.ip_assign:
-  linode.cloud.ip_assign:
     region: us-east
     assignments:
      - address: 0.0.0.0
--- a/plugins/modules/ip_assign.py
+++ b/plugins/modules/ip_assign.py
@@ -65,7 +65,10 @@ class Module(LinodeModuleBase):
     """Module for assigning IPs to Linodes in a given Region"""
     def __init__(self) -> None:
         self.module_arg_spec = SPECDOC_META.ansible_spec
-        self.results = {"changed": False}
+        self.results = {
+            "changed": False,
+            "actions": [],
+        }
         super().__init__(module_arg_spec=self.module_arg_spec)
 
     def flatten_ips(self, ips):
@@ -90,17 +93,17 @@ class Module(LinodeModuleBase):
                     return self.results
 
             self.client.networking.ips_assign(region, *assignments)
+            self.register_action(f"IP assignments completed: {assignments}")
 
             for assignment in assignments:
                 linode = Instance(self.client, assignment["linode_id"])
                 linode._api_get()
                 if assignment["address"] not in self.flatten_ips(linode.ips):
                     self.fail(msg=f"IP assignments not changed: {assignments}")
+                    self.results["changed"] = False
                     return self.results
         except Exception as exc:
             self.fail(msg=f"failed to set IP assignments {assignments}: {exc}")
-
-        self.results["changed"] = True
 
         return self.results
 
--- a/plugins/modules/ip_assign.py
+++ b/plugins/modules/ip_assign.py
@@ -61,8 +61,10 @@ SPECDOC_META = SpecDocMeta(
     return_values={},
 )
 
+
 class Module(LinodeModuleBase):
     """Module for assigning IPs to Linodes in a given Region"""
+
     def __init__(self) -> None:
         self.module_arg_spec = SPECDOC_META.ansible_spec
         self.results = {
@@ -73,8 +75,10 @@ class Module(LinodeModuleBase):
 
     def flatten_ips(self, ips):
         """Flatten a linodes IPs to quickly check the assignment"""
-        addrs = [v.address for v in ips.ipv4.public
-                  + ips.ipv4.private + ips.ipv4.reserved]
+        addrs = [
+            v.address
+            for v in ips.ipv4.public + ips.ipv4.private + ips.ipv4.reserved
+        ]
         addrs += [v.range for v in ips.ipv6.ranges]
         return addrs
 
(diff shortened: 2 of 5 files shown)


plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,100 @@
...
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for ip_assign module"""
+        params = filter_null_values(self.module.params)
+        assignments = params.get("assignments")
+        region = params.get("region")
+
+        try:
+            self.client.networking.ips_assign(region, *assignments)
+        except Exception as exc:
+            self.fail(msg=f"failed to set IP assignments {assignments}: {exc}")
```
lgarber-akamai, line 87: Could we check to see if the assignments have already been made so `changed` will only be set to true if the assignments have been updated?

lgarber-akamai (changes requested):
Nice work so far!

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,100 @@
...
+    return_values={},
+)
+
+
+class Module(LinodeModuleBase):
+    """Module for assigning IPs to Linodes in a given Region"""
+
+    def __init__(self) -> None:
+        self.module_arg_spec = SPECDOC_META.ansible_spec
+        self.results = {"changed": False}
```
yec-akamai, line 70: Should we initialize `actions` as well?

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,100 @@
...
+
+        super().__init__(
+            module_arg_spec=self.module_arg_spec,
+            required_one_of=[],
+            mutually_exclusive=[],
```
yec-akamai, line 75: If we don't need to specify `required_one_of` and `mutually_exclusive` in a module, is it okay to just remove them here or we have to pass in the empty array?

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,100 @@
...
+        params = filter_null_values(self.module.params)
+        assignments = params.get("assignments")
+        region = params.get("region")
+
+        try:
+            self.client.networking.ips_assign(region, *assignments)
+        except Exception as exc:
+            self.fail(msg=f"failed to set IP assignments {assignments}: {exc}")
+
+        self.results["changed"] = True
```
yec-akamai, line 89: Should we call `self.register_action(...)` to register the action?

tests/integration/targets/ip_assign/tasks/main.yml
```diff
@@ -0,0 +1,52 @@
...
+      linode.cloud.instance_info:
+        id: '{{create_instance_2.instance.id }}'
+      register: updated_instance_2
+
+    - name: Assert changes
+      assert:
+        that:
+          - updated_instance.networking.ipv4.public[0].address == create_instance_2.networking.ipv4.public[0].address
+          - updated_instance_2.networking.ipv4.public[0].address == create_instance.networking.ipv4.public[0].address
+
```
yec-akamai, line 47: We probably can add a block like this to destroy the testing instances in the end:
```
  always:
    - ignore_errors: true
      block:
        - linode.cloud.instance:
            label: '{{ instance_create.instance.label }}'
            state: absent
```

plugins/module_utils/doc_fragments/ip_assign.py
```diff
@@ -0,0 +1,11 @@
+"""Documentation fragments for the ip_assign module"""
+specdoc_examples = ['''
+- name: Assign IP to Linode
+  linode.cloud.ip_assign:
```
lgarber-akamai, line 4: Looks like this line is a duplicate

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,123 @@
...
+
+        try:
+            for assignment in assignments:
+                linode = Instance(self.client, assignment["linode_id"])
+                linode._api_get()
+                if assignment["address"] in self.flatten_ips(linode.ips):
+                    self.fail(
+                        msg=f"IP: {assignment['address']} already assigned to instance: {assignment['linode_id']}"
+                    )
+                    return self.results
```
lgarber-akamai, line 99: We can return without the failure here so it'll just show up as unchanged in the playbook run 👍

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,123 @@
...
+            self.client.networking.ips_assign(region, *assignments)
+            self.register_action(f"IP assignments completed: {assignments}")
+
+            for assignment in assignments:
+                linode = Instance(self.client, assignment["linode_id"])
+                linode._api_get()
+                if assignment["address"] not in self.flatten_ips(linode.ips):
+                    self.fail(msg=f"IP assignments not changed: {assignments}")
+                    self.results["changed"] = False
+                    return self.results
```
lgarber-akamai, line 110: Since we're checking that the IPs were already assigned before calling `ips_assign`, I think this block can be dropped

lgarber-akamai (changes requested):
Great work so far! Just a few small changes

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,111 @@
...
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for ip_assign module"""
+        params = filter_null_values(self.module.params)
+        assignments = params.get("assignments")
+        region = params.get("region")
+
+        try:
+            for assignment in assignments:
+                linode = Instance(self.client, assignment["linode_id"])
+                linode._api_get()
```
lgarber-akamai, line 94: nitpick: This request might be unnecessary since `.ips` makes a network request to a separate endpoint

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,111 @@
...
+                if assignment["address"] in self.flatten_ips(linode.ips):
+                    return self.results
```
lgarber-akamai, line 96: I think this snippet will return early if any of the assignments already exist rather than only if all of the assignments already exist.

lgarber-akamai (changes requested):
Looks great! Just one more small thing and this should be good to go

plugins/modules/ip_assign.py
jriddle-linode: Since IP Assign only executes if the entire order can be fulfilled i dont think returning early is a problem.

plugins/modules/ip_assign.py
```diff
@@ -0,0 +1,110 @@
...
+    ),
+    "region": SpecField(
+        type=FieldType.string,
+        required=True,
+        description=["The Region to operate in."],
+    ),
+}
+
+SPECDOC_META = SpecDocMeta(
+    description=["Assign IPs to Linodes in a given Region."],
```
lgarber-akamai, line 56: Just a reminder that we'll want a warning here as discussed on Friday.

lgarber-akamai (approved):
LGTM! Just needs a `make gendocs` run and a few small changes to fix the linting errors 👍

plugins/modules/ip_assign.py
zliang-akamai: But it can still be partially executed if returning early, right?

zliang-akamai:
Test run: https://github.com/linode/ansible_linode/actions/runs/6177081277

zliang-akamai (approved):

PR #404, opened 2023-08-30
new: Add VPC and VPC Subnet modules

lgarber-akamai:
This change adds modules for the VPC and VPCSubnet API objects, as well as a new `ListModuleBase` helper class to significantly reduce the amount of boilerplate needed when creating list modules. In the future we could likely implement something similar for info modules.

This PR does **not** include the corresponding interface-related changes.

In order to test this change, you will need to install `linode_api4-python` from the [proj/vpc](https://github.com/linode/linode_api4-python/tree/proj/vpc) branch.

New modules:
- `vpc`
- `vpc_info`
- `vpc_list`
- `vpc_subnet`
- `vpc_subnet_info`
- `vpc_subnet_list`

```
export TEST_API_URL=https://...
export TEST_CA_FILE=...

make TEST_ARGS="-v vpc_basic" test
make TEST_ARGS="-v vpc_subnet" test
```

### DIFF
--- a/plugins/module_utils/linode_common_list.py
+++ b/plugins/module_utils/linode_common_list.py
@@ -1,0 +1,130 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""This module allows users to list SSH keys in their Linode profile."""
+
+from __future__ import absolute_import, division, print_function
+
+from typing import Any, Dict, Optional
+
+from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
+    LinodeModuleBase,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import (
+    global_authors,
+    global_requirements,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import (
+    construct_api_filter,
+    get_all_paginated,
+)
+from ansible_specdoc.objects import (
+    FieldType,
+    SpecDocMeta,
+    SpecField,
+    SpecReturnValue,
+)
+
+spec_filter = {
+    "name": SpecField(
+        type=FieldType.string,
+        required=True,
+        description=[
+            "The name of the field to filter on.",
+        ],
+    ),
+    "values": SpecField(
+        type=FieldType.list,
+        element_type=FieldType.string,
+        required=True,
+        description=[
+            "A list of values to allow for this field.",
+            "Fields will pass this filter if at least one of these values matches.",
+        ],
+    ),
+}
+
+
+class ListModuleBase(LinodeModuleBase):
+    """Module for getting a list of SSH keys in the Linode profile"""
+
+    display_name = ""
+    result_field = ""
+    endpoint = ""
+    docs_url = ""
+    examples = []
+    response_samples = []
+
+    def __init__(self) -> None:
+        self.module_arg_spec = self.spec.ansible_spec
+        self.results: Dict[str, Any] = {self.result_field: []}
+
+        super().__init__(module_arg_spec=self.module_arg_spec)
+
+    def exec_module(self, **kwargs: Any) -> Optional[dict]:
+        """Entrypoint for list module"""
+
+        filter_dict = construct_api_filter(self.module.params)
+
+        self.results[self.result_field] = get_all_paginated(
+            self.client,
+            self.endpoint,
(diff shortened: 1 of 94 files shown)


plugins/module_utils/linode_common_list.py
```diff
@@ -0,0 +1,157 @@
...
+    docs_url = ""
+    parents: List[ListModuleParent] = []
+    examples = []
+    response_samples = []
+
+    def __init__(self) -> None:
+        self.module_arg_spec = self.spec.ansible_spec
+        self.results: Dict[str, Any] = {self.result_field: []}
+
+        super().__init__(module_arg_spec=self.module_arg_spec)
```
zliang-akamai, line 76: I think we can initialize the attributes in `__init__()` function and let the sub-classes call the parent's `__init__()`. I feel by this way, we would align with the typical object-oriented design pattern.

plugins/module_utils/linode_helper.py
```diff
@@ -305,3 +305,33 @@ def poll_condition(
...
+
+def get_resource_safe(list_func: Callable[[], List[Any]]) -> Any:
+    """
+    Wraps a resource list function with error handling.
+    If no entries are returned, this function returns None rather than
+    raising an error.
+    """
+    try:
+        list_results = list_func()
+        return None if len(list_results) < 1 else list_results[0]
```
zliang-akamai, line 318: Would it be fine to let the caller of this function to check whether the list is empty or not? I think empty list won't raise an error, right?

plugins/module_utils/linode_helper.py
```diff
@@ -305,3 +305,33 @@ def poll_condition(
...
+    except Exception as exception:
+        raise Exception(f"failed to get resource: {exception}") from exception
```
zliang-akamai, line 320:
```suggestion
    return get_resource_safe_condition(list_results, lambda _: True)
```

zliang-akamai:
Others look good to me!

plugins/module_utils/linode_common_list.py
zliang-akamai:
```python
class Parent:
    a: int
    b: str
    def __init__(self, a, b):
        self.a = a
        self.b = b

class Child1(Parent):
    def __init__(self, a, b):
        super().__init__(a, b)

class Child2(Parent):
    def __init__(self):
        super().__init__(123, "abc")
```

Just an example to clarify my thought.

Another advantage I can think of is it will let the developer of the sub-class be aware of all the attributes and ensure each attribute is considered, and people may be confusing about how the attributes work in the declarative sub-class.

And feel free to resolve this is we are not planning it in this PR.

zliang-akamai (approved):

plugins/module_utils/linode_common_list.py
lgarber-akamai: Ahh I see, thanks for the feedback! We can discuss this further in the standalone PR.

zliang-akamai:
But the CI fails?

lgarber-akamai:
@zliang-akamai The CI failure is due to the non-project branch version of `linode_api4` being installed

plugins/modules/vpc_subnet.py
```diff
@@ -0,0 +1,152 @@
...
+        params = filter_null_values(
+            {k: v for k, v in self.module.params.items() if k in CREATE_FIELDS}
+        )
+
+        try:
+            return vpc.subnet_create(**params)
+        except Exception as exception:
+            return self.fail(msg="failed to create VPC: {0}".format(exception))
+
+    def _update(self, subnet: VPCSubnet) -> None:
```
yec-akamai, line 97: Should we be able to update VPC subnet `label`? https://docs.google.com/document/d/1BE8LkLRzP2dY4S8eUuW1hWHLyJmUWh7PZSOqE94amno/edit#heading=h.nhueb1yp6ywu

yec-akamai (commented):
Great work! Just a small comment.

plugins/modules/vpc_subnet.py
yec-akamai: Gotcha. Thanks for the explanation!

yec-akamai (approved):
Nice work! 🎉

zliang-akamai (approved):

PR #409, opened 2023-09-13
new: Add InfoModule class to simplify info module implementations

lgarber-akamai:
## 📝 Description

This change adds a shared `InfoModuleBase` class that allows for the easy creation of info modules with very little boilerplate.

This change also switches from directly calling `pytest` to using the `ansible-test units` wrapper.

## ✔️ How to Test

```
make TEST_ARGS="-v instance_basic" test
make TEST_ARGS="-v stackscript_info" test
```

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -8,9 +8,14 @@ name: Run Unit test
 jobs:
   run-tests:
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: .ansible/collections/ansible_collections/linode/cloud
     steps:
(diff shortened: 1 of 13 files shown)


amisiorek-akamai:
Tests pass locally with the exception of one, just want to double check that this is outside our scope.

`TASK [instance_basic : Create a Linode instance with additional ips and without a root password] ***
fatal: [testhost]: FAILED! => {"changed": false, "msg": "Error from Linode API: [400] Additional IPv4 addresses require technical justification.  Please contact support describing your requirement"}`

amisiorek-akamai (approved):

PR #411, opened 2023-09-18
new: adding `firewall_id` field to nodebalancers

jriddle-linode:
## 📝 Description

Adding `firewall_id` field to creating nodebalancers.

## ✔️ How to Test

`make TEST_ARGS='-v nodebalancer_firewall' test`

### DIFF
--- a/plugins/modules/nodebalancer.py
+++ b/plugins/modules/nodebalancer.py
@@ -28,7 +28,7 @@ from ansible_specdoc.objects import (
     SpecField,
     SpecReturnValue,
 )
-from linode_api4 import NodeBalancer, NodeBalancerConfig, NodeBalancerNode
+from linode_api4 import NodeBalancer, NodeBalancerConfig, NodeBalancerNode, Firewall
 
 linode_nodes_spec = {
     "label": SpecField(
@@ -347,13 +347,10 @@ class LinodeNodeBalancer(LinodeModuleBase):
         label = params.get("label")
         region = params.get("region")
         firewall_id = params.get("firewall_id")
-
         try:
             return self.client.nodebalancer_create(region, label=label, firewall_id=firewall_id)
         except Exception as exception:
-            return self.fail(
-                msg="failed to create nodebalancer: {0}".format(exception)
-            )
+            return self.fail(msg=f"failed to create nodebalancer: {exception}")
 
     def _create_config(
         self, node_balancer: NodeBalancer, config_params: dict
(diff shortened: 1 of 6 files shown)


plugins/modules/nodebalancer.py
```diff
@@ -342,9 +346,10 @@ def _create_nodebalancer(self) -> Optional[NodeBalancer]:
         params = self.module.params
         label = params.get("label")
         region = params.get("region")
+        firewall_id = params.get("firewall_id")
 
         try:
-            return self.client.nodebalancer_create(region, label=label)
+            return self.client.nodebalancer_create(region, label=label, firewall_id=firewall_id)
```
lgarber-akamai, line 352: We'll probably want to implement update or validation logic to prevent users from updating the `firewall_id` and expecting that a change has been applied.

This flow could look something like:
1. If `firewall_id` isn't specified, skip
2. Get the firewall that `firewall_id` is pointing to, raise an error if it doesn't exist
3. Check that there is a device under the firewall pointing to the NodeBalancer, if not raise a "non-updatable" error

lgarber-akamai (changes requested):

plugins/modules/nodebalancer.py
```diff
@@ -523,6 +520,14 @@ def _update_nodebalancer(self) -> None:
         if "configs" in params.keys():
             params.pop("configs")
 
+        if "firewall_id" in params.keys():
+            firewall_id = params.pop("firewall_id")
+            fw = Firewall(self.client, firewall_id)
+            if not fw:
+                self.fail(f"firewall: {firewall_id} doesn't exist")
+            if params.get('label') not in [x.id for x in fw.devices if x.entity is NodeBalancer]:
+                self.fail(f"firewall: {firewall_id} does not contain this nodebalancer")
```
lgarber-akamai, line 529: Could we adjust this message to say something like `The firewall_id for a NodeBalancer cannot currently be updated after creation.`?

plugins/modules/nodebalancer.py
```diff
@@ -523,6 +520,14 @@ def _update_nodebalancer(self) -> None:
         if "configs" in params.keys():
             params.pop("configs")
 
+        if "firewall_id" in params.keys():
+            firewall_id = params.pop("firewall_id")
+            fw = Firewall(self.client, firewall_id)
```
lgarber-akamai, line 525: I don't think this will make an API request immediately due to lazy-loading, so you might need to use `self.client.load(...)` instead

lgarber-akamai (changes requested):
Looks good! Just one small thing and this should be ready to go

lgarber-akamai (approved):
LGTM! Nice work

amisiorek-akamai (approved):
LGTM!

PR #418, opened 2023-09-27
Support specifying a custom CA path

amisiorek-akamai:
## 📝 Description

Users can now specify a custom CA file by passing in its path through a ca_path argument for each module.

### DIFF
--- a/plugins/module_utils/linode_common.py
+++ b/plugins/module_utils/linode_common.py
@@ -80,6 +80,10 @@ LINODE_COMMON_ARGS = {
         "doc_hide": True,
         "fallback": (env_fallback, ["LINODE_UA_PREFIX"]),
     },
+    "ca_path": {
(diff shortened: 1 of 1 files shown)


plugins/module_utils/linode_common.py
```diff
@@ -80,6 +80,10 @@
         "doc_hide": True,
         "fallback": (env_fallback, ["LINODE_UA_PREFIX"]),
     },
+    "ca_path": {
+        "type": "str",
+        "description": "A path to a custom certificate authority for using alternate APIs.",
```
lgarber-akamai, line 85: Could there also be an environment variable fallback like we do for `linode_token`? e.g. `LINODE_CA`

lgarber-akamai (approved):
Looks good, nice work!

zliang-akamai (approved):

PR #422, opened 2023-10-03
Adding testing for LinodeBaseModule

amisiorek-akamai:
## 📝 Description

Adds unit testing for LinodeBaseModule and migrates all existing unittest to pytest.

## ✔️ How to Test

`make unittest`

### DIFF
--- a/tests/unit/module_utils/test_linode_common.py
+++ b/tests/unit/module_utils/test_linode_common.py
@@ -1,0 +1,21 @@
+from types import SimpleNamespace
+
+from unittest import TestCase
+import os
+
+from ansible_collections.linode.cloud.tests.unit.base import TestModuleBase
+
+class TestLinodeModuleBase(TestCase):
+
+    def setUp(self, **kwargs):
+        mock_module = TestModuleBase()
+        mock_module.module = SimpleNamespace(params=kwargs)
+
+        return mock_module
+
+    def test_module_ca_path_override(self):
+        os.environ['LINODE_CA'] = "env_ca"
+        mock_module = self.setUp(api_token="testing", api_version=None, api_url="/", ua_prefix=None, ca_path="foobar")
+
+        client = mock_module.client
+        assert client.ca_path == "foobar"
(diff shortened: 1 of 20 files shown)


tests/unit/module_utils/test_linode_common.py
```diff
@@ -0,0 +1,21 @@
...
+        mock_module.module = SimpleNamespace(params=kwargs)
+
+        return mock_module
+
+    def test_module_ca_path_override(self):
+        os.environ['LINODE_CA'] = "env_ca"
+        mock_module = self.setUp(api_token="testing", api_version=None, api_url="/", ua_prefix=None, ca_path="foobar")
+
+        client = mock_module.client
+        assert client.ca_path == "foobar"
```
zliang-akamai, line 21: The `setUp` method is intended for automatically invocation from the Python unittest framework before running all test cases in the class. `pytest` is used by `ansible-test` to test execute our unit tests, so I think it would be nicer if we can use a simple and straightforward `pytest` style to write the test case.

```suggestion
def test_module_ca_path_override(self):
    os.environ["LINODE_CA"] = "env_ca"
    mock_module = TestModuleBase()
    mock_module.module = SimpleNamespace(
        params={
            "api_token": "testing",
            "api_version": None,
            "api_url": "/",
            "ua_prefix": None,
            "ca_path": "foobar",
        }
    )
    client = mock_module.client
    assert client.ca_path == "foobar"
```

amisiorek-akamai: Looks like we consume unittest in most of the unit tests for Ansible...do you think we should migrate to exclusively use pytest?

tests/unit/module_utils/test_linode_common.py
```diff
@@ -0,0 +1,21 @@
+from types import SimpleNamespace
+import os
+
+from ansible_collections.linode.cloud.tests.unit.base import TestModuleBase
+
+class TestLinodeModuleBase():
```
zliang-akamai, line 6: Just a minor optional suggestion, if the test case is not depends on any class/object attributes (`self.some_attr`), we can even remove the classes and only keep the functions, lol

lgarber-akamai (approved):
Tests pass on my end, nice work!

PR #442, opened 2023-11-21
fix failing test and skip DBaaS test

vshanthe:
## 📝 Description
Address failing integration tests
Skip DBaaS tests by default

Fixing the failing test case in integration  
## ✔️ How to Test
make TEST_ARGS="-v user_grants" test
export RUN_LONG_TESTS=1
make test
## 📷 Preview

**If applicable, include a screenshot or code snippet of this change. Otherwise, please remove this section.**


.github/workflows/integration-tests.yml
```diff
@@ -43,6 +43,12 @@ jobs:
         run: rm -rf ~/.ansible/test && mkdir -p ~/.ansible/test && ssh-keygen -m PEM -q -t rsa -N '' -f ~/.ansible/test/id_rsa
 
       - name: run tests
-        run: make testall
+        run: |
+          if [ "$RUN_LONG_TESTS" = "1" ]; then
+            make testall
+          else
+            echo "Skipping long tests."
+          fi
```
zliang-akamai, line 51: We would still want to run not so long tests in else, right?

.github/workflows/integration-tests.yml
```diff
@@ -43,6 +43,12 @@ jobs:
...
         env:
           LINODE_API_TOKEN: ${{ secrets.DX_LINODE_TOKEN }}
+          RUN_LONG_TESTS: ${{ env.RUN_LONG_TESTS }}
```
zliang-akamai, line 54: I wonder how can we configure `RUN_LONG_TESTS`..

.github/workflows/integration-tests.yml
vshanthe: Good to have if we opt to skip ryt

zliang-akamai: But in the else branch here, it only echo a string?

vshanthe: - name: run tests
  run: |
    if [ "$RUN_LONG_TESTS" = "1" ]; then
      make testall
    else
      echo "Skipping long tests."
      make testall
  env:
    LINODE_API_TOKEN: ${{ secrets.DX_LINODE_TOKEN }}
    RUN_LONG_TESTS: ${{ secrets.RUN_LONG_TESTS }}

zliang-akamai: I think in the workflow we can always skip these tests? 

`make testall` always run all tests, I think.
Can we find a way to avoid running the long tests while still going to run the non-long tests?

tests/integration/targets/user_grants/tasks/main.yaml
```diff
@@ -71,7 +71,7 @@
     - name: Assert grants updated
       assert:
         that:
-          - no_changes.changed == False
+          - no_changes.changed == True
```
lgarber-akamai, line 74: The test step this assertion references doesn't seem to change anything, so we might need to investigate what is causing the diff here :+1:

vshanthe: actions is consider as "Updated grants" in both Update and No changes 
 and the changed status will have the value 'true" 
 we can have verify in other way like below 
       assert:
        that:
          - no_changes.grants.global.add_linodes == update_grants.grants.global.add_linodes

Makefile
```diff
@@ -59,12 +59,20 @@ gendocs:
 	python scripts/render_readme.py $(COLLECTION_VERSION)
 
 integration-test: create-integration-config
+ifdef RUN_LONG_TESTS
 	ansible-test integration $(TEST_ARGS)
+else
+	ansible-test integration $(TEST_ARGS) --exclude-tag longtests
```
lgarber-akamai, line 65: Could this tag be applied to the DBaaS test cases?

tests/integration/targets/user_grants/tasks/main.yaml
```diff
@@ -71,7 +71,7 @@
     - name: Assert grants updated
       assert:
         that:
-          - no_changes.changed == False
```
zliang-akamai, line 74: Is this needed for skipping the long tests?

zliang-akamai (changes requested):
https://github.com/linode/ansible_linode/actions/runs/7043293716/job/19168814987
The integration tests failed to stared..

tests/integration/targets/user_grants/tasks/main.yaml
lgarber-akamai: This is to address an unrelated test failure

vshanthe:
> https://github.com/linode/ansible_linode/actions/runs/7043293716/job/19168814987 The integration tests failed to stared..

> https://github.com/linode/ansible_linode/actions/runs/7056305624/job/19208064609

zliang-akamai:
My `ansible-test` doesn't recognize `--exclude-tag` argument.. Does it work on your end?

Do you have a version number of it that has this feature?

And I can't see any `longtests` tag is applied to the test case, liked [mentioned here](https://github.com/linode/ansible_linode/pull/442#discussion_r1408367162), can you add them to the specific test cases that is long which we would want to skip?

vshanthe:
> My `ansible-test` doesn't recognize `--exclude-tag` argument.. Does it work on your end?
> 
> Do you have a version number of it that has this feature?
> 
> And I can't see any `longtests` tag is applied to the test case, liked [mentioned here](https://github.com/linode/ansible_linode/pull/442#discussion_r1408367162), can you add them to the specific test cases that is long which we would want to skip?

> i have added the tag and can be performed by below command 
>  make EXCLUDE_TAG="test_name" test

.github/workflows/integration-tests.yml
zliang-akamai: Hi @vshanthe I think we will still want to run some tests even if the long tests are skipped

zliang-akamai:
Hi @vshanthe, so one of the goals of this is to run all the tests without running the long tests in the GitHub workflow, instead of skipping all the tests.. Can you double check the unsolved comments above and ensure it's working?

.github/workflows/integration-tests.yml
```diff
@@ -45,4 +45,4 @@ jobs:
       - name: run tests
         run: make testall
         env:
-          LINODE_API_TOKEN: ${{ secrets.DX_LINODE_TOKEN }}
+          LINODE_API_TOKEN: ${{ secrets.DX_LINODE_TOKEN }}
```
zliang-akamai, line 48: Can we have a blank line at the end of the file to fix the GitHub warning?
image

zliang-akamai (approved):
Tested locally, observed DBaaS tests are skipped by default, and `user_grants` test has been passed. Nice work!

ykim-1 (approved):
LGTM, can you include a full integration test execution via GHA to verify everything is working as expected?

lgarber-akamai (approved):
Tested locally and everything works as expected, excellent work!

`make TEST_ARGS="-v mysql_basic" test` - Ran no tests
`make TEST_ARGS="-v mysql_basic --tags never" test` - Ran the mysql_basic test

PR #447, opened 2023-12-11
new: Add `account_availability_list` module

yec-akamai:
## 📝 Description

Add a new list module `account_availability_list` to show a list of region availabilities to the account. This module is currently not filterable because the API doesn't support filtering.

## ✔️ How to Test

To test again Alpha, you probably need to overwrite the following variables:

```
TEST_API_URL=alpha/api-url
TEST_API_CA=your-local-ca-file
LINODE_TOKEN=alpha-linode-token
```

`make TEST_ARGS="-v account_availability_list" test`

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,7 @@ Modules for retrieving and filtering on multiple Linode resources.
 
 Name | Description |
 --- | ------------ |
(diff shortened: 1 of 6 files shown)


vshanthe (approved):

zliang-akamai (approved):
Tested and worked well on my end!

PR #452, opened 2023-12-20
new: Improve InfoModule attribute spec generation

lgarber-akamai:
## 📝 Description

This change makes the following improvements to the attribute spec generation in the InfoModule base class:
- Individual info module attributes are now treated as required.
- Info module attributes now are now marked as in-conflict with other attributes. 

### Unit Testing

```
make unittest
```

### Manually Testing Required Individual Attributes

1. Inside of an ansible_linode sandbox environment, run the following playbook:

```yaml
---
- name: Required attribute testing
  hosts: localhost
  tasks:
    - name: Get info about a type without specifying an ID
      linode.cloud.type_info:
```

2. Observe the following error is raised:

```
missing required arguments: id
```

3. Update the playbook to include a type ID:

```yaml
---
- name: Required attribute testing
  hosts: localhost
  tasks:
    - name: Get info about a type without specifying an ID
      linode.cloud.type_info:
        id: g6-nanode-1
```

4. Observe the playbook passes as expected.

### DIFF
--- a/tests/unit/module_utils/test_linode_common_info.py
+++ b/tests/unit/module_utils/test_linode_common_info.py
@@ -9,7 +9,7 @@ from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import gl
 
 class TestLinodeInfoModule:
 
-    @pytest.fixture
+    @pytest.fixture(scope="function")
     def mock_module(self):
         return InfoModule(
             primary_result=InfoModuleResult(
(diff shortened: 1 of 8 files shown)


tests/unit/module_utils/test_linode_common_info.py
```diff
@@ -9,7 +9,7 @@
 
 class TestLinodeInfoModule:
 
-    @pytest.fixture
+    @pytest.fixture(scope="function")
```
lgarber-akamai, line 12: Added the function scope here just for some extra confidence that the below tests will not conflict.

zliang-akamai (approved):
Took some time to understand how everything work, lol. Everything looks good, and local test passed!

jriddle-linode (approved):
LGTM after we found the dependency issue.

PR #459, opened 2024-01-09
fix: Resolve issue with filtering in `object_cluster_info` module

lgarber-akamai:
## 📝 Description

This change moves all filterable fields in `object_cluster_info` module from API filtering to client-side filtering. This is necessary because the Linode API does not allow filtering on any fields when listing clusters.

This change should resolve the test failures for the `object_basic` test case but NOT the `object_cluster_list` test case. Due to the potential implementation complexity of client-side filtering in a list module and the fact that OBJ clusters will soon be deprecated, we decided against applying a similar fix to `object_cluster_list`.

**NOTE: Tests were previously passing because the first returned cluster was always `us-east-1`, but as far as I can tell this module never worked as expected.**

## ✔️ How to Test

Integration Testing:

```
make TEST_ARGS="-v object_basic" test
```

### DIFF
--- a/tests/integration/targets/object_cluster_list/tasks/main.yaml
+++ b/tests/integration/targets/object_cluster_list/tasks/main.yaml
@@ -1,24 +1,27 @@
 - name: object_cluster_list
   block:
-    - name: List object clusters with no filter
-      linode.cloud.object_cluster_list:
-      register: no_filter
+    - debug:
+        msg: Skipping...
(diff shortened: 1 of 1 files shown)


zliang-akamai (approved):
Tested and worked well!

yec-akamai (approved):
Nice work! Tests passed locally 🎉

PR #476, opened 2024-03-14
doc: Add disclaimer to documentation for modules that consume beta endpoints

lgarber-akamai:
## 📝 Description

This change adds a disclaimer to the documentation for modules that consume beta (`v4beta`) endpoints. I intentionally opted not to add `api_version` validation logic to prevent any testing/edge case issues.

**NOTE: We do not implicitly override the API version because the default behavior for beta usage should be opt-out.**

## 📷 Preview

ansible_linode_docs_modules_ip_share_md_at_doc_beta-endpoints_·_lgarber-akamai_ansible_linode


yec-akamai (approved):
Nice work!

zliang-akamai (approved):
LGTM!

Issue #419, opened 2023-09-29
[Bug]: ModuleNotFoundError: No module named 'polling'


ajprok:
### Ansible Version

ansible [core 2.15.4]   config file = /home/ajit/current-projects/lap/lap-infrastructure/configuration/ansible.cfg   configured module search path = ['/home/ajit/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']   ansible python module location = /home/ajit/.local/lib/python3.10/site-packages/ansible   ansible collection location = /home/ajit/.ansible/collections:/usr/share/ansible/collections   executable location = /home/ajit/.local/bin/ansible   python version = 3.10.12 (main, Jun 11 2023, 05:26:28) [GCC 11.4.0] (/home/ajit/current-projects/lap/venv/bin/python3)   jinja version = 3.1.2   libyaml = True

### Ansible Configuration

```
CONFIG_FILE() = /home/ajit/current-projects/lap/lap-infrastructure/configuration/ansible.cfg
DEFAULT_HOST_LIST(/home/ajit/current-projects/lap/lap-infrastructure/configuration/ansible.cfg) = ['/home/ajit/current-projects/lap/lap-infrastructure/configuration/inventory']
DEFAULT_ROLES_PATH(/home/ajit/current-projects/lap/lap-infrastructure/configuration/ansible.cfg) = ['/home/ajit/current-projects/lap/lap-infrastructure/configuration/roles']
HOST_KEY_CHECKING(/home/ajit/current-projects/lap/lap-infrastructure/configuration/ansible.cfg) = False
```

### Effected Components

linode.cloud.firewall_info

### Expected Behavior

I'm using terraform to run an ansible command after instance creation. The ansible command is being run in a virtual environment (**venv**). Following is the relevant code:
```
provisioner "local-exec" {
    working_dir = "${path.root}/../../configuration"
    command = <<EOT
      # Activate venv
      source ../../venv/bin/activate
      # Ansible command
      # Trailing comma in inventory required
      ansible-playbook -i '${self.ip_address},' -e 'gateway=true' -e 'cloud_provider=${var.cloud_provider}' -e 'instance_details=${jsonencode(self)}' init.yml > temp.log
    EOT
  }
```
The output of `pip freeze` is given below:
```
ansible==8.4.0
ansible-core==2.15.4
ansible-specdoc==0.0.14
appdirs==1.4.4
apt-xapian-index==0.49
argcomplete==3.1.2
Babel==2.8.0
[...]

PR #477, opened 2024-03-14
doc: Add API token requirement in module description

yec-akamai:
## 📝 Description

Add to each module description that `LINODE_API_TOKEN` is required to be set.

Addressed #419

### DIFF
--- a/docs/modules/account_availability_info.md
+++ b/docs/modules/account_availability_info.md
@@ -2,8 +2,6 @@
 
 Get info about a Linode Account Availability.
 
-LINODE_API_TOKEN environment variable is required.
-
 - [Examples](#examples)
 - [Parameters](#parameters)
 - [Return Values](#return-values)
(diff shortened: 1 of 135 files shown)


zliang-akamai:
I think user may also specify `api_token` parameter in a task level?

jriddle-linode (approved):

yec-akamai:
> I think user may also specify a `api_token` parameter in a task level?

Yes you're correct. I modify the change to make `api_token` visible in the parameters, and also highlight the usage of environment variable.

zliang-akamai:
Since all `api_token` `SpecField`s are basically same, can we define it from a single place and import it to everywhere else?

Maybe in the `module_utils` module?
```python
SpecField(
    type=FieldType.string,
    required=False,
    description="The Linode account personal access token. "
    "It is necessary to run the module. "
    "It can be exposed by the "
    "environment variable `LINODE_API_TOKEN` instead.",
)
```

yec-akamai:
> Thanks for adding it, and it looks nice! Since all `api_token` `SpecField`s are basically same, can we define it from a single place and import it to everywhere else?
> 
> Maybe in the `module_utils` module?
> 
> ```python
> SpecField(
>     type=FieldType.string,
>     required=False,
>     description="The Linode account personal access token. "
>     "It is necessary to run the module. "
>     "It can be exposed by the "
>     "environment variable `LINODE_API_TOKEN` instead.",
> )
> ```

It's a good idea! I'll make the change

zliang-akamai (approved):
Nice work!

PR #479, opened 2024-03-21
new: Add `child_account_info` and `child_account_list` modules; add user_type to user response examples

lgarber-akamai:
## 📝 Description

This change adds the `child_account_info` and `child_account_list` modules, as well as a new `user_type` attribute to the user-related examples.

NOTE: This PR intentionally does not include functionality for creating child account tokens because they are short-lived and we cannot view their state after creation. If you feel this should be added to the scope of this PR, definitely let me know 🙂 

For the time being users can use something like this:

```yaml
- linode.cloud.api_request:
    path: 'account/child-accounts/{{ euuid }}/token'
    method: POST
```

## ✔️ How to Test

The following test steps assume you have pulled this PR locally and run `pip install --force -r requirements.txt`.

Additionally, the following environment variables should be set before running any of the following:

```bash
export TEST_API_CA=$PWD/cacert.pem
export TEST_API_URL=https://.../
export LINODE_TOKEN=...
```

### Integration Testing

```bash
make TEST_ARGS="-v child_accounts --tags=parent_child" test
make TEST_ARGS="-v user_basic" test
make TEST_ARGS="-v user_list" test
```

### Manual Testing

```yaml
---
  hosts: localhost
  tasks:
    - name: List all children under the current account
      linode.cloud.child_account_list: { }
      register: children

    - name: Get information about a specific Child Account
      linode.cloud.child_account_info:
        euuid: '{{ children.child_accounts[0].euuid }}'
      register: child

    - name: All child accounts
      debug:
        var: children.child_accounts

    - name: Single child account
      debug:
        var: child.child_account
```

2. Ensure the playbook succeeds as expected.

### DIFF
--- a/tests/integration/targets/child_accounts/tasks/main.yaml
+++ b/tests/integration/targets/child_accounts/tasks/main.yaml
@@ -1,0 +1,31 @@
+- name: child_accounts
+  tags:
+    - never
+    - parent_child
+  block:
+    - name: List all children under the current account
+      linode.cloud.child_account_list: {}
+      register: children
+
+    - assert:
+        that:
+          - children.child_accounts | length > 0
+          - children.child_accounts[0].email is not none
+          - children.child_accounts[0].euuid is not none
+
+    - name: Get information about a specific Child Account
+      linode.cloud.child_account_info:
+        euuid: '{{ children.child_accounts[0].euuid }}'
+      register: child
+
+    - assert:
+        that:
+          - child.child_account.email == children.child_accounts[0].email
+          - child.child_account.euuid == children.child_accounts[0].euuid
+
+  environment:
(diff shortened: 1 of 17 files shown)


tests/integration/targets/child_accounts/tasks/main.yaml
```diff
@@ -0,0 +1,31 @@
+- name: child_accounts
+  tags:
+    - never
+    - parent_child
```
lgarber-akamai, line 4: This makes sure we don't run this test alongside the normal suite because it requires a special testing environment.

zliang-akamai (approved):
Works well on my end. Nice work!

yec-akamai (approved):
Great work! Tests works well on my end

PR #481, opened 2024-03-22
Add `normalize_api_url` Helper Method

zliang-akamai:
## 📝 Description

Reduced pain during configuring API URL

## ✔️ How to Test
You can try any of the tests, e.g. `make TEST_ARGS="-v account_availability_info" test `

### DIFF
--- a/plugins/module_utils/linode_common.py
+++ b/plugins/module_utils/linode_common.py
@@ -250,13 +250,23 @@ class LinodeModuleBase:
                 "with id {resource_id}: {exception}"
             )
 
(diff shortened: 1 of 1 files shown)


lgarber-akamai (commented):
@zliang-akamai I wonder if it'd make sense to add this to the Python SDK so we can reuse it across Python projects 👀

zliang-akamai:
@lgarber-akamai this would break the unit test pattern in the Python SDK. Let me close this and think about if it's possible to integrate it into the SDK.

PR #492, opened 2024-04-09
new: Add handling to update logic for unordered lists; use update helper in NB & Instance modules

lgarber-akamai:
## 📝 Description

This pull request adds logic to the `handle_updates` helper function to respect lists annotated as unordered during the diffing process. This ensures unordered fields like `tags` will not trigger infinite diffs with equivalent values.

Additionally, this PR moves the basic update logic in the `instance` and `nodebalancer` modules over to the `handle_updates` helper. These are NOT the only modules that can be moved over to the helper but I figure we can do that work in a separate PR to keep this PR tightly scoped. (I added an entry for this in the internal feedback tracker)

## ✔️ How to Test

Additionally you will need to run `pip install --force -r requirements.txt` in order to install from the `dev` branch of the Python SDK.

### Integration Testing

```
make test TEST_ARGS="-v nodebalancer_reordered_tags"
```

### Manual Testing

1. In an Ansible Collection sandbox environment (e.g. dx-devenv), run the following playbook in verbose (-v) mode:

```yaml
- name: test
  hosts: localhost
  tasks:
    - set_fact:
        r: "{{ 1000000000 | random }}"

    - name: Create a NodeBalancer with tags
      linode.cloud.nodebalancer:
        label: 'ansible-test-{{ r }}'
        region: us-ord
        tags:
          - test1
          - test2
          - test3
        state: present
      register: create_nodebalancer

    - name: Reorder the NodeBalancer's tags
      linode.cloud.nodebalancer:
        label: 'ansible-test-{{ r }}'
        region: us-ord
        tags:
          - test3
          - test1
          - test2
        state: present
      register: reorder_nodebalancer_tags

    - name: Update the NodeBalancer's tags
      linode.cloud.nodebalancer:
        label: 'ansible-test-{{ r }}'
        region: us-ord
        tags:
          - test1
          - test4
          - test5
          - test6
        state: present
      register: update_nodebalancer_tags
```

2. Ensure the playbook runs without any errors.
3. Ensure the `Reorder the NodeBalancer's tags` task is marked as unchanged (green).
4. Ensure the `Update the NodeBalancer's tags` task is marked as changed (yellow) and reflects the diff in the `actions` field.

### DIFF
--- a/plugins/module_utils/linode_helper.py
+++ b/plugins/module_utils/linode_helper.py
@@ -116,6 +116,9 @@ def handle_updates(
 
     obj._api_get()
 
+    # We need the type to access property metadata
+    property_metadata = type(obj).properties
+
     # Update mutable values
     params = filter_null_values(params)
 
@@ -128,7 +131,15 @@ def handle_updates(
 
         old_value = parse_linode_types(getattr(obj, key))
 
-        if new_value != old_value:
+        has_diff = new_value != old_value
+
+        # We should convert properties to sets
+        # if they are annotated as unordered in the
+        # Python SDK.
+        if property_metadata.get(key).unordered:
+            has_diff = set(old_value) != set(new_value)
+
+        if has_diff:
             if key in mutable_fields:
                 put_request[key] = new_value
                 result.add(key)
--- a/plugins/modules/nodebalancer.py
+++ b/plugins/modules/nodebalancer.py
@@ -9,7 +9,6 @@ import copy
 from typing import Any, List, Optional, Set, Tuple, cast
 
 import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.nodebalancer as docs
-import linode_api4
 from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
     LinodeModuleBase,
 )
@@ -20,6 +19,7 @@ from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import (
 from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import (
     dict_select_matching,
     filter_null_values,
+    handle_updates,
     paginated_list_to_json,
 )
 from ansible_specdoc.objects import (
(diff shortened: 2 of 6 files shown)


requirements.txt
```diff
@@ -1,4 +1,7 @@
-linode-api4>=5.13.1
+# TODO: Blocked by SDK release; revert before merging to dev
+# linode-api4>=5.13.1
+git+https://github.com/linode/linode_api4-python@dev
```
lgarber-akamai, line 3: The unordered annotation is a super recent addition so we'll need to cut a new Python SDK release before merging this PR

plugins/module_utils/linode_helper.py
```diff
@@ -128,7 +131,15 @@ def handle_updates(
...
         old_value = parse_linode_types(getattr(obj, key))
 
-        if new_value != old_value:
+        has_diff = new_value != old_value
+
+        # We should convert properties to sets
+        # if they are annotated as unordered in the
+        # Python SDK.
+        if property_metadata.get(key).unordered:
+            has_diff = set(old_value) != set(new_value)
```
lgarber-akamai, line 140: Apparently the API request function doesn't like JSON serializing sets so I couldn't mutate old_value and new_value here

lgarber-akamai:
Integration test suite run: https://github.com/linode/ansible_linode/actions/runs/8620714908

plugins/modules/nodebalancer.py
```diff
@@ -248,6 +248,12 @@
         type=FieldType.integer,
         description=["The ID of the Firewall to assign this NodeBalancer to."],
     ),
+    "tags": SpecField(
```
lgarber-akamai, line 251: For some reason the tags fields was missing from the spec here

jriddle-linode:
It'll be `improvement` once we merge to main.

yec-akamai (approved):
Nice work! Everything passed locally. The new handle_updates function really makes the code clear!

jriddle-linode (approved):
LGTM passes locally with forced requirements.

PR #493, opened 2024-04-16
Add `networking_info` attribute to inventory instances

### LINKED ISSUE #451: [Feature]: IPv6 Range in Inventory
### Description

It would be awesome if an additional IPv6 range allocated to a Linode would be returned when using the inventory. Currently it only looks like `ipv4` and `ipv6` address are returned, which is from the `instance` section of the `instance_info` response

Looking at `linode.cloud.instance_info` it returns a `networking` section, so perhaps the ideal implementation is to have a variable in the inventory plugin configuration that would allow the user to enable returning this `networking` block or not (default being false), eg

```
plugin: linode.cloud.instance
...
networking: true
```

This could be extended as well for other blocks returned in `instance_info` like

```
disks: true
configs: true
```

Example of this for another inventory is [netbox inventory](https://github.com/netbox-community/ansible_modules/blob/devel/plugins/inventory/nb_inventory.py#L102-L107), it allows users to pull in extra information like interfaces, eg of each device

### New/Affected Components

Inventory plugin

### Example Ansible Config

My use-case would be something like using it in configuring the interfaces on a Linode in `/etc/network/interfaces`

```
interface lo:1 inet static
  address {{ networking.ipv6.global.range | ansible.utils.nthhost(1) }}/128
```

zliang-akamai:
## 📝 Description

closes #451

This is to add `networking_info` with data received from the [Networking Information List API](https://www.linode.com/docs/api/linode-instances/#networking-information-list) to instances in the dynamic inventory.

### Automated Testing

**Automated testing doesn't work on macOS**, and you may want to test it in a Linode.

```bash
ansible-test integration -v instance_inventory
```

### Manual Testing

- `make install`
- Create a dynamic inventory file called `myinstance.linode.yaml` with the content below.

```yaml
plugin: linode.cloud.instance
```

- Create a Linode instance.
- Execute `ansible-inventory --list -i /path/to/myinstance.linode.yaml -v` and verify `networking_info` attribute is in the output Linode instances.

### DIFF
--- a/plugins/inventory/instance.py
+++ b/plugins/inventory/instance.py
@@ -19,7 +19,7 @@ from ansible_collections.linode.cloud.plugins.module_utils.linode_common import 
 )
 from linode_api4.objects import Instance
 
-DOCUMENTATION = r"""
+DOCUMENTATION = """
     name: instance
     author:
       - Luke Murphy (@decentral1se)
@@ -59,7 +59,7 @@ DOCUMENTATION = r"""
           type: list
 """
 
-EXAMPLES = r"""
+EXAMPLES = """
 # Minimal example. `LINODE_API_TOKEN` is exposed in environment.
 plugin: linode.cloud.instance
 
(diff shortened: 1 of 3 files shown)


lgarber-akamai (approved):

yec-akamai (approved):
Works well locally! Nice work

zliang-akamai:
Blocked by https://github.com/linode/linode_api4-python/pull/399

PR #395, opened 2023-07-27
fix: Resolve division-by-zero error for empty list responses

lgarber-akamai:
## 📝 Description

This change fixes a division-by-zero error that would occur when no entries are returned from a list module.

## ✔️ How to Test

```
make TEST_ARGS="-v image_list" test
```

PR #494, opened 2024-04-16
new: Add support to `available` field to account_availability info and list module

yec-akamai:
## 📝 Description

Update documentations to include the new field `available` to account_availability info and list module. Add test case to verify the new field works on ansible.

## ✔️ How to Test

The endpoints are currently available on alpha and the PR depends on the [python SDK change #395](https://github.com/linode/linode_api4-python/pull/395). You'll need to pull down the python SDK PR and install it locally to run the test cases. You'll also need to set up test url and token to test in alpha.

```
TEST_API_URL=alpha/api-url
TEST_API_CA=your-local-ca-file
LINODE_TOKEN=alpha-linode-token
```
```
make TEST_ARGS="-v account_availability_list" test 
make TEST_ARGS="-v account_availability_info" test 
```

Manual test:
1. In a sandbox environment, run the following ansible playbook script:
```ansible-playbook
  hosts: localhost
  tasks:
    - name: Get account availability information
      linode.cloud.account_availability_info:
        region: us-east
      register: account_availability

    - name: Output info about the accout availability
      debug:
        msg: |
          Account Availability Info:
            Region: {{ account_availability.account_availability.region }}
            Unavailable: {{ account_availability.account_availability.unavailable }}
            Available: {{ account_availability.account_availability.available }}
```
2. Observe that the result correctly shows up.

### DIFF
--- a/tests/integration/targets/account_availability_info/tasks/main.yaml
+++ b/tests/integration/targets/account_availability_info/tasks/main.yaml
@@ -9,6 +9,7 @@
       assert:
         that:
           - account_availability.account_availability.region == "us-east"
+          - account_availability.account_availability.available | length + account_availability.account_availability.unavailable | length == 4
 
   environment:
     LINODE_UA_PREFIX: '{{ ua_prefix }}'
--- a/docs/modules/account_availability_info.md
+++ b/docs/modules/account_availability_info.md
@@ -40,6 +40,7 @@ Get info about a Linode Account Availability.
         
         {
           "region": "us-east",
+          "available": ["NodeBalancers", "Block Storage", "Kubernetes"],
           "unavailable": ["Linode"]
         }
         
--- a/docs/modules/account_availability_list.md
+++ b/docs/modules/account_availability_list.md
@@ -48,10 +48,12 @@ List and filter on Account Availabilitys.
         [
             {
               "region": "ap-west",
+              "available": ["NodeBalancers", "Block Storage", "Kubernetes"],
               "unavailable": ["Linode"]
             },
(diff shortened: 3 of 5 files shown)


tests/integration/targets/account_availability_info/tasks/main.yaml
```diff
@@ -9,6 +9,7 @@
       assert:
         that:
           - account_availability.account_availability.region == "us-east"
+          - account_availability.account_availability.available | length + account_availability.account_availability.unavailable | length == 4
```
ykim-1, line 12: Would it worth it to also check for the actual content? e.g. ["Linodes", "NodeBalancers", "Block Storage", "Kubernetes"]

lgarber-akamai: I don't think we can guarantee the length of the availabilities either, so it might be make sense to check if certain entries exist in the fields rather than explicitly checking the list as a whole :+1:

lgarber-akamai (approved):
LGTM

zliang-akamai (approved):
LGTM!

tests/integration/targets/account_availability_info/tasks/main.yaml
yec-akamai: Good point! The content of available and unavailable may change due to data center capability, so it may break our test if we check the content. However, the "total" will always be the four services: Linodes, NodeBalancers, Block Storage, Kubernetes. At least this won't change based on the current design ;) I've added an comment to describe it.

PR #498, opened 2024-04-18
Fix plural noun in docs

zliang-akamai:
## 📝 Description
Push the naming responsibility to individual module to fix the plural noun in the doc.

## ✔️ How to Test

You can just read the docs.


jriddle-linode (approved):

lgarber-akamai (approved):
Nice catch, thanks for fixing this!

PR #499, opened 2024-04-18
Add VPC IP List Modules

zliang-akamai:
## 📝 Description

Built on top of https://github.com/linode/ansible_linode/pull/498, thus the diff will be smaller after merging that PR.

This is to add `all_vpcs_ip_list` and `vpc_ip_list` modules. I realized this is easier to manage than combine them because they are separated API endpoints. But this is definitely debatable., so free feel to left you opinion.

## ✔️ How to Test
```bash
make TEST_ARGS="-v vpc_ip_list" test
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,7 +86,7 @@ Modules for retrieving and filtering on multiple Linode resources.
 
 Name | Description |
 --- | ------------ |
-[linode.cloud.account_availability_list](./docs/modules/account_availability_list.md)|List and filter on Account Availabilitys.|
+[linode.cloud.account_availability_list](./docs/modules/account_availability_list.md)|List and filter on Account Availabilities.|
 [linode.cloud.database_engine_list](./docs/modules/database_engine_list.md)|List and filter on Managed Database engine types.|
 [linode.cloud.database_list](./docs/modules/database_list.md)|List and filter on Linode Managed Databases.|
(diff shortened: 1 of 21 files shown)


plugins/modules/all_vpcs_ip_list.py
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""This module contains all of the functionality for listing Linode VPCs."""
```
lgarber-akamai, line 4: I think these docstrings need to be updated 👍

README.md
```diff
@@ -87,6 +87,7 @@ Modules for retrieving and filtering on multiple Linode resources.
 Name | Description |
 --- | ------------ |
 [linode.cloud.account_availability_list](./docs/modules/account_availability_list.md)|List and filter on Account Availabilities.|
+[linode.cloud.all_vpcs_ip_list](./docs/modules/all_vpcs_ip_list.md)|List and filter on all VPC IP Addresses.|
```
lgarber-akamai, line 90: If possible, it'd be nice to have this module name start with `vpc` to group it with the rest of the vpc-related docs.

Maybe something like `vpc_ips_all_list`?

jriddle-linode (approved):
Works locally, LGTM after Lena's feedback.

lgarber-akamai (approved):
Thank you for adjusting the module name! LGTM and all tests are passing locally

PR #503, opened 2024-05-06
new: Add modules for Placement group

yec-akamai:
## 📝 Description

This PR includes the following change for VM placement functionalities:
1. Add new modules:
    - placement_group: accept both label and id as identifier
    - placement_group_info
    - placement_group_list

2. In the `instance` module, allow to create a new linode instance under a placement group.

## ✔️ How to Test

Run `pip install --force -r requirements.txt`  to install required dependencies. 

Also, you need to set up environment variables to test against alpha:

```
export TEST_API_CA=$PWD/cacert.pem
export TEST_API_URL=https://.../
export LINODE_TOKEN=...
```

**Integration tests:**
```
make TEST_ARGS="-v placement_group" test
make TEST_ARGS="-v placement_group_info" test
make TEST_ARGS="-v placement_group_list" test
make TEST_ARGS="-v instance_basic" test
```

**Manual Test:**

1. In a sandbox environment, i.e. dx-devenv, run the following ansible playbook
```yaml
  hosts: localhost
  tasks:
    - name: Create a Linode placement group
      linode.cloud.placement_group:
        label: my-ansible-test-pg
        region: us-east
        affinity_type: anti_affinity:local
        state: present
      register: pg_created

    - name: Get a Linode placement group info
      linode.cloud.placement_group_info:
        id: '{{ pg_created.placement_group.id }}'
      register: pg

    - name: Update a Linode placement group label
      linode.cloud.placement_group:
        id: '{{ pg.placement_group.id }}'
        label: '{{ pg.placement_group.label }}-updated'
        state: present
      register: pg_updated

    - name: List Linode placement groups for the current account
      linode.cloud.placement_group_list:
        filters:
          - name: label
            values: ['{{ pg_updated.placement_group.label }}']
      register: pg_list
```

2. Observe no error is raised.

3. Create an instance under the placement group
```yaml
    - name: Create a Linode instance under a placement group
      linode.cloud.instance:
        label: 'ansible-test-pg-{{ r }}'
        type: g6-nanode-1
        region: us-east
        placement_group:
          id: '{{ pg_updated.placement_group.id }}'
          compliant_only: false
        state: present
      register: instance_pg
```

4. Observe that the instance is created successfully.

5. Clean up the resources created.
```yaml
    - name: Delete a Linode instance
      linode.cloud.instance:
        label: '{{ instance_pg.instance.label }}'
        state: absent
      register: delete_instance_pg

    - name: Delete a placement group
      linode.cloud.placement_group:
        label: '{{ pg_updated.placement_group.label }}'
        state: absent
      register: pg_deleted
```

### DIFF
--- a/plugins/modules/instance.py
+++ b/plugins/modules/instance.py
@@ -1207,6 +1207,7 @@ class LinodeInstance(LinodeModuleBase):
                 "backups_enabled",
                 "type",
                 "region",
+                "placement_group",
             )
         }
 
--- a/plugins/modules/instance.py
+++ b/plugins/modules/instance.py
@@ -1240,6 +1240,19 @@ class LinodeInstance(LinodeModuleBase):
                     "non-updatable field".format(self._instance.label)
                 )
 
+        pg = params.get("placement_group")
+        if pg is not None:
+            if pg.get("id") != self._instance.placement_group.id:
+                self.fail(
+                    "failed to update instance {0}: placement_group.id is a "
+                    "non-updatable field".format(self._instance.label)
+                )
+            if pg.get("compliant_only"):
+                self.warn(
+                    "placement_group.compliant_only is non-updatable and only can be "
+                    "specified in the instance creation."
+                )
+
         # Update interfaces
         self._update_interfaces()
 
--- a/plugins/modules/instance.py
+++ b/plugins/modules/instance.py
@@ -301,6 +301,20 @@ spec_additional_ipv4 = {
     )
 }
 
+linode_instance_placement_group_spec = {
+    "id": SpecField(
+        type=FieldType.integer,
+        description="The id of the placement group.",
+        required=True,
+    ),
+    "compliant_only": SpecField(
+        type=FieldType.bool,
+        description="Whether the newly added/migrated/resized linode "
+        "must be compliant for flexible placement groups.",
+        default=False,
+    ),
+}
+
 linode_instance_spec = {
     "label": SpecField(
         type=FieldType.string,
@@ -499,6 +513,11 @@ linode_instance_spec = {
             "Tags are for organizational purposes only.",
         ],
         editable=True,
+    ),
+    "placement_group": SpecField(
+        type=FieldType.dict,
+        suboptions=linode_instance_placement_group_spec,
+        description=["A Placement Group to create this Linode under."],
     ),
 }
 
--- a/plugins/modules/placement_group.py
+++ b/plugins/modules/placement_group.py
@@ -30,9 +30,12 @@ from ansible_specdoc.objects import (
 from linode_api4 import PlacementGroup
 
 placement_group_spec = {
+    "id": SpecField(
+        type=FieldType.integer,
+        description="The unique ID of the placement group.",
+    ),
     "label": SpecField(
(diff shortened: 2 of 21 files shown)


plugins/modules/placement_group.py
```diff
@@ -0,0 +1,179 @@
...
+            docs_url="TBD",
+            type=FieldType.dict,
+            sample=docs.result_placement_group_samples,
+        )
+    },
+)
+
+CREATE_FIELDS = {"label", "region", "affinity_type", "is_strict"}
+# Fields that can be updated on an existing placement group
+MUTABLE_FIELDS = {"label"}
```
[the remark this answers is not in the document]
yec-akamai: Makes sense! I'm wondering if we should allow user to use the PG `id` as the identifier, then they can modify label in that case?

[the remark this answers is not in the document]
lgarber-akamai: That's a good idea! I'm not sure if we have a precedent for allowing users to specify an ID but I don't think it'd hurt to implement it here 👍

plugins/modules/instance.py
```diff
@@ -500,6 +514,11 @@
         ],
         editable=True,
     ),
+    "placement_group": SpecField(
+        type=FieldType.dict,
+        suboptions=linode_instance_placement_group_spec,
+        description=["A Placement Group to create this Linode under."],
+    ),
 }
```
lgarber-akamai, line 522: It looks like re-running a playbook with an instance assigned to a PG and no changes gives the following error:

```
{"changed": false, "msg": "RuntimeError: failed to update PlacementGroup: 1946 -> {'id': 1946, 'compliant_only': False}: placement_group is a non-updatable field"}
```

We may need to introduce some logic in the update function to make sure it doesn't attempt to diff when the PG is unchanged 👀

yec-akamai: Good catch 👍 I excluded placement_group from the update_params to fix this issue.

lgarber-akamai: Thanks for fixing this! Looks like there are some backend issues right now but I'll test the fix as soon as they're resolved

plugins/modules/instance.py
```diff
@@ -1220,6 +1240,19 @@ def _update_instance(self) -> None:
...
                 )
 
+        pg = params.get("placement_group")
+        if pg is not None:
+            if pg.get("id") != self._instance.placement_group.id:
+                self.fail(
+                    "failed to update instance {0}: placement_group.id is a "
+                    "non-updatable field".format(self._instance.label)
+                )
+            if pg.get("compliant_only"):
```
yec-akamai, line 1250: @lgarber-akamai The value of `compliant_only` is not returned from API and we don't have a state like tf to find its old value, so I'm thinking returning a warning here to remind user that compliant_only is not updatable.

lgarber-akamai: Sounds good to me!

lgarber-akamai (approved):
All E2E test are passing and everything works as expected in my manual testing, great work!

zliang-akamai (approved):
LGTM, tested and worked well. Nice work!

PR #505, opened 2024-05-08
fix: Add conditional update logic for the `ip_rdns` module

### LINKED ISSUE #500: [Bug]: linode.cloud.rdns applies change when rdns is already correctly set
### Ansible Version

ansible [core 2.16.5]

### Ansible Configuration

_No response_

### Effected Components

linode.cloud.ip_rdns (version 0.25.6)

Run the same ip_rdns task multiple times.  It reports "changed" each time.

### Expected Behavior

If should report "ok" whenever the task is run again (when the rdns is already correctly set).

linode.cloud.ip_rdns should return `{changed: false}` when the ip rdns is already correctly set.

### Actual Behavior

Subsequent runs continue to report "changed":

```
TASK [Rdns ipv4] *****************************************************************************************************************************************************************************************************
changed: [localhost] => {"actions": ["Updated reverse DNS of the IP address *redacted_ip* to be *redacted_fqdn*"], "changed": true, "ip": {"address": "*redacted_ip*", "gateway": "*redacted*", "linode_id": *redacted*, "prefix": 24, "public": true, "rdns": "*redacted_fqdn*", "region": "*redacted*", "subnet_mask": "255.255.255.0", "type": "ipv4", "vpc_nat_1_1": null}}
```

### Steps to Reproduce

Run the same ip_rdns task multiple times.  It reports "changed" every time.  The expected behavior is that it only report "changed" the first time only.

lgarber-akamai:
## 📝 Description

This pull request adds conditional update logic for the `ip_rdns` module, which prevents the result from being marked as `changed` when the RDNS of an IP has not been updated.

Resolves #500 

## ✔️ How to Test

### Integration Testing

```
make TEST_ARGS="-v ip_rdns" test
make TEST_ARGS="-v ip_rdns_ipv6" test
```

### Manual Testing

```yaml
  hosts: localhost
  tasks:
      linode.cloud.instance:
        label: 'rdns-test'
        type: g6-nanode-1
        region: us-mia
        state: present
      register: instance_create

    - set_fact:
        new_rdns: '{{ instance_create.instance.ipv4[0] }}.nip.io'
        old_rdns: '{{ instance_create.instance.ipv4[0] | replace(".", "-") }}.ip.linodeusercontent.com'
        ip_address: '{{ instance_create.instance.ipv4[0] }}'

    - name: Update reverse DNS of the instance's primary IPv4 address
      linode.cloud.ip_rdns:
        address: '{{ ip_address }}'
        rdns: '{{ new_rdns }}'
        state: present
      register: rdns_update

    - name: Assert that the RDNS was successfully updated
      assert:
        that:
          - rdns_update.changed
          - rdns_update.ip.address == ip_address
          - rdns_update.ip.rdns == new_rdns

    - name: Make no changes to the RDNS address
      linode.cloud.ip_rdns:
        address: '{{ ip_address }}'
        rdns: '{{ new_rdns }}'
        state: present
      register: rdns_update_no_changes

    - name: Assert that the RDNS address was unchanged
      assert:
        that:
          - not rdns_update_no_changes.changed
          - rdns_update_no_changes.ip.address == ip_address
          - rdns_update_no_changes.ip.rdns == new_rdns

    - name: Revert the RDNS address to the Linode API default
      linode.cloud.ip_rdns:
        address: '{{ ip_address }}'
        state: absent
      register: rdns_revert

    - name: Assert that the RDNS was successfully reverted for the instance
      assert:
        that:
          - rdns_revert.changed
          - rdns_revert.ip.address == ip_address
          - rdns_revert.ip.rdns == old_rdns

    - name: Attempt to revert the RDNS again
      linode.cloud.ip_rdns:
        address: '{{ ip_address }}'
        state: absent
      register: rdns_revert_unchanged

    - name: Assert that the RDNS was unchanged
      assert:
        that:
          - not rdns_revert_unchanged.changed
          - rdns_revert_unchanged.ip.address == ip_address
          - rdns_revert_unchanged.ip.rdns == old_rdns
```
2. Ensure all assertions pass successfully.
3. Ensure all log outputs match the expected behavior.

### DIFF
--- a/plugins/module_utils/linode_common.py
+++ b/plugins/module_utils/linode_common.py
@@ -247,7 +247,7 @@ class LinodeModuleBase:
             )
             return self.fail(
                 msg=f"failed to get {resource_name} "
-                "with id {resource_id}: {exception}"
+                f"with id {resource_id}: {exception}"
             )
 
     @property
--- a/plugins/modules/ip_rdns.py
+++ b/plugins/modules/ip_rdns.py
@@ -5,7 +5,7 @@
 
 from __future__ import absolute_import, division, print_function
 
-from typing import Any, Optional
+from typing import Any, Optional, Union
 
 import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.ip_info as ip_docs
 import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.ip_rdns as ip_rdns_docs
@@ -40,7 +40,7 @@ ip_rdns_spec = {
     "rdns": SpecField(
         type=FieldType.string,
         editable=True,
-        description=["The desired rDNS value."],
+        description=["The desired RDNS value."],
     ),
 }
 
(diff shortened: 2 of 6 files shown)


plugins/module_utils/linode_common.py
```diff
@@ -247,7 +247,7 @@ def _get_resource_by_id(
             )
             return self.fail(
                 msg=f"failed to get {resource_name} "
-                "with id {resource_id}: {exception}"
+                f"with id {resource_id}: {exception}"
```
lgarber-akamai, line 250: Unrelated issue I noticed in the error message for 404s 👀

plugins/modules/ip_rdns.py
```diff
@@ -82,28 +82,53 @@ def __init__(self) -> None:
...
+        Returns whether the old RDNS value and the proposed new RDNS for the
+        IP address differ.
+        """
+
+        # If the RDNS address is being reset, compare the old RDNS against
+        # the Linode API default
+        if isinstance(new_rdns, ExplicitNullValue):
+            ip_address = self.module.params.get("address")
+            new_rdns = (
+                f"{ip_address.replace('.', '-')}.ip.linodeusercontent.com"
```
lgarber-akamai, line 98: I'm not the biggest fan of hardcoding the default RDNS address but I'm not aware of a better way to work around it as of now

yec-akamai: It seems the way we have to work around the default RDNS. Another thing I can think of is makeing this logic into a helper function like `build_default_rdns(...)`, may help a bit with readability.

zliang-akamai (approved):
Looks good and runs well! Thanks for fixing it!

plugins/modules/ip_rdns.py
lgarber-akamai: I like the idea of a helper function, I'll push up a commit to fix that 👍

lgarber-akamai:
I added IPv6 RDNS diffing support as a part of the new helper since its default is a bit different than the IPv4 default. Sorry for the re-requested review @zliang-akamai !

yec-akamai (approved):
Nice work! All tests passed locally 🎉

ykim-1 (approved):
LGTM, thanks for fixing this

PR #507, opened 2024-05-20
test: Fix broken assertion in instance_interfaces_vpc test

lgarber-akamai:
## 📝 Description

This pull request adjusts an assertion in the `instance_interfaces_vpc` test to reference the second interface returned from `subnet.linodes.*.interfaces`, which is necessary because all interfaces (including non-VPC interfaces) are returned for this field.

## ✔️ How to Test

### Integration Tests

```
make TEST_ARGS="-v instance_interfaces_vpc" test
```


jriddle-linode (approved):
Works locally!

yec-akamai (approved):

PR #517, opened 2024-06-06
fix: Resolve issues in firewall diff logic

lgarber-akamai:
## 📝 Description

This pull request resolves the following issues present in the `firewall` module's diff logic:
* Diffs are always detected on the `rules` field due to a new `fingerprint` field returned by the API
* An error is raised when attempting to update devices because the `devices` field was not excluded from the key/value update logic

## ✔️ How to Test

### Integration Testing

```
make TEST_ARGS="-v firewall_basic firewall_device firewall_icmp firewall_list firewall_update" test
```

### Manual Testing

```yaml
---
  hosts: localhost
  tasks:
  - name: Create an Instance
    linode.cloud.instance:
      label: my-instance
      type: g6-nanode-1
      region: us-mia
      state: present
    register: instance

  - name: Create a Linode Firewall
    linode.cloud.firewall:
      label: "my-firewall"
      rules:
        inbound_policy: DROP
        inbound:
          - label: allow-http-in
            addresses:
              ipv6: ["ff00::/8"]
            ports: "80,443"
            protocol: TCP
            action: ACCEPT

        outbound_policy: DROP
        outbound: []
      state: present

  - name: Refresh the firewall
    linode.cloud.firewall:
      label: "my-firewall"
      rules:
        inbound_policy: DROP
        inbound:
          - label: allow-http-in
            addresses:
              ipv6: ["ff00::/8"]
            ports: "80,443"
            protocol: TCP
            action: ACCEPT

        outbound_policy: DROP
        outbound: []
      state: present
    register: refresh

  - name: Assert that the firewall rules have not been updated
    assert:
      that:
        - not refresh.changed

  - name: Update the firewall's devices
    linode.cloud.firewall:
      label: "my-firewall"
      devices:
        - id: "{{ instance.instance.id }}"
          type: linode
      rules:
        inbound_policy: DROP
        inbound:
          - label: allow-http-in
            addresses:
              ipv6: [ "ff00::/8" ]
            ports: "80,443"
            protocol: TCP
            action: ACCEPT

        outbound_policy: DROP
        outbound: [ ]
      state: present
    register: update_devices

  - name: Assert that the firewall's devices have been updated
    assert:
      that:
        - update_devices.changed
        - update_devices.devices[0].entity.id == instance.instance.id
```
2. Ensure all assertions pass and the playbook finished running successfully.

### DIFF
--- a/plugins/modules/firewall.py
+++ b/plugins/modules/firewall.py
@@ -153,6 +153,11 @@ linode_firewall_spec: dict = {
         editable=True,
         description=["The status of this Firewall."],
     ),
+    "tags": SpecField(
+        type=FieldType.list,
+        editable=True,
+        description=["A list of tags to apply to this Firewall."]
+    ),
     "state": SpecField(
         type=FieldType.string,
         description=["The desired state of the target."],
--- a/plugins/modules/firewall.py
+++ b/plugins/modules/firewall.py
@@ -355,17 +355,20 @@ class LinodeFirewall(LinodeModuleBase):
     def _change_rules(self) -> Optional[dict]:
         """Updates remote firewall rules relative to user-supplied new rules,
         and returns whether anything changed."""
-        local_rules = filter_null_values_recursive(self.module.params["rules"])
-        remote_rules = filter_null_values_recursive(
-            mapping_to_dict(self._firewall.rules)
+        local_rules = (
+            filter_null_values_recursive(self.module.params["rules"]) or {}
+        )
+        remote_rules = (
+            filter_null_values_recursive(mapping_to_dict(self._firewall.rules))
+            or {}
         )
 
-        # user did not specify any rules updates
-        if local_rules is None:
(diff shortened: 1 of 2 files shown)


plugins/modules/firewall.py
```diff
@@ -153,6 +153,11 @@
         editable=True,
         description=["The status of this Firewall."],
     ),
+    "tags": SpecField(
+        type=FieldType.list,
+        editable=True,
+        description=["A list of tags to apply to this Firewall."]
+    ),
```
lgarber-akamai, line 160: This isn't directly related to this fix but it's a gap I noticed that shouldn't cause any functional changes 🙂

ykim-1 (approved):
Thanks for fixing it quickly! Tests pass locally

zliang-akamai (approved):
Works well on my end. Nice work!

PR #519, opened 2024-06-11
new: Add support for LKE cluster control plane ACLs

lgarber-akamai:
## 📝 Description

This pull request adds support for configuring the ACL for an LKE cluster's control plane using the `lke_cluster` module.

## ✔️ How to Test

The following test steps assume you have pulled down this PR locally, run `make install`, and configured your Linode account according to the description of TPT-2676.

### Integration Testing

```
make TEST_ARGS="-v lke_cluster_basic" test
make TEST_ARGS="-v lke_cluster_acl" test
make TEST_ARGS="-v lke_cluster_info_ro" test
```

### Limited Access Integration Testing

1. Configure your Linode account to NOT have access to LKE ACLs.
2. Run the following integration tests:

```
make TEST_ARGS="-v lke_cluster_basic" test
make TEST_ARGS="-v lke_cluster_info_ro" test
```

### Manual Testing

1. In an Ansible Collection sandbox environment (e.g. dx-devenv), run the following playbook in verbose mode:

```yaml
---
  hosts: localhost
  tasks:
    - name: Resolve all available K8s versions
      linode.cloud.lke_version_list: {}
      register: get_lke_versions

    - set_fact:
        lke_versions: '{{ get_lke_versions.lke_versions|sort(attribute="id") }}'

    - name: Create a Linode LKE cluster
      linode.cloud.lke_cluster:
        label: "lke-acl-test"
        region: us-mia
        k8s_version: "{{ lke_versions[0].id }}"
        node_pools:
          - type: g6-standard-1
            count: 1
        acl:
          enabled: true
          addresses:
            ipv4:
              - 0.0.0.0/0
            ipv6:
              - 2001:db8:1234:abcd::/64
        skip_polling: true
        state: present

    - name: Get info about the cluster
      linode.cloud.lke_cluster_info:
        label: "lke-acl-test"
```

2. Ensure the playbook runs successfully and the resulting cluster's ACL configuration matches the expected ACL configuration.
3. Make arbitrary changes to the ACL configuration in this playbook.
4. Re-run the playbook and ensure all functionality works as expected.

### DIFF
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -175,7 +175,10 @@ linode_lke_cluster_spec = {
         type=FieldType.dict,
         suboptions=linode_lke_cluster_acl,
         editable=True,
-        description=["The ACL configuration for this cluster's control plane."],
+        description=[
+            "The ACL configuration for this cluster's control plane.",
+            "NOTE: Control Plane ACLs may not currently be available to all users.",
+        ],
     ),
     "node_pools": SpecField(
         editable=True,
@@ -365,11 +368,21 @@ class LinodeLKECluster(LinodeModuleBase):
         Handles the update logic for an LKE cluster's control plane ACL configuration.
         """
         control_plane_acl = self._safe_get_cluster_acl(cluster)
-        configured_acl = self.module.params.get("acl")
+        configured_acl = copy.deepcopy(self.module.params.get("acl"))
 
         # We don't want to make any changes if the user has not explicitly defined an ACL
         if configured_acl is None:
             return
+
+        # [] and null are equivalent values for address fields,
+        # so we need to account for this when diffing
+        configured_addresses = configured_acl.get("addresses")
+        if configured_addresses is not None:
+            ipv4 = configured_addresses.get("ipv4")
+            ipv6 = configured_addresses.get("ipv6")
+
+            configured_acl["addresses"]["ipv4"] = None if ipv4 == [] else ipv4
+            configured_acl["addresses"]["ipv6"] = None if ipv6 == [] else ipv6
 
         user_defined_keys = set(linode_lke_cluster_acl.keys())
         current_acl = (
@@ -579,9 +592,9 @@ class LinodeLKECluster(LinodeModuleBase):
     def _populate_dashboard_url_poll(self, cluster: LKECluster) -> None:
         def condition() -> bool:
             try:
-                self.results[
-                    "dashboard_url"
-                ] = cluster.cluster_dashboard_url_view()
+                self.results["dashboard_url"] = (
+                    cluster.cluster_dashboard_url_view()
+                )
             except ApiError as error:
                 if error.status != 503:
                     raise error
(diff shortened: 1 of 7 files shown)


plugins/modules/lke_cluster.py
```diff
@@ -290,6 +362,45 @@ def _create_cluster(self) -> Optional[LKECluster]:
...
+
+        user_defined_keys = set(linode_lke_cluster_acl.keys())
+        current_acl = (
+            control_plane_acl.dict if control_plane_acl is not None else {}
+        )
+
+        # Only diff on keys that can be defined by the user
+        current_acl = {
+            k: v for k, v in current_acl.items() if k in user_defined_keys
+        }
```
lgarber-akamai, line 394: This diff logic is super messy but I can't think of a better way to implement it. Any feedback would be much appreciated!

plugins/modules/lke_cluster.py
```diff
@@ -290,6 +360,43 @@ def _create_cluster(self) -> Optional[LKECluster]:
...
+        user_defined_keys = set(linode_lke_cluster_acl.keys())
+        current_acl = control_plane_acl if control_plane_acl is not None else {}
+
+        # Only diff on keys that can be defined by the user
+        current_acl = {
+            k: v for k, v in current_acl.items() if k in user_defined_keys
+        }
+
+        if configured_acl == current_acl:
+            return
```
lgarber-akamai, line 393: This diff logic is super messy but I can't think of a better way to implement it. Any feedback would be much appreciated!

plugins/modules/lke_cluster_info.py
```diff
@@ -143,14 +143,34 @@ def _get_cluster_from_kwargs(self, **kwargs: Any) -> LKECluster:
 
         return self.fail(msg="one of `label` or `id` must be specified")
 
+    def _safe_get_cluster_acl(
```
lgarber-akamai, line 146: Given this is used in both `lke_cluster` and `lke_cluster_info`, I'm thinking it might make sense to move it into the shared helper package 👍

plugins/modules/lke_cluster.py
```diff
@@ -290,6 +341,43 @@ def _create_cluster(self) -> Optional[LKECluster]:
...
```
lgarber-akamai, line 374: This diff logic is super messy but I can't think of a better way to implement it. Any feedback would be much appreciated!

ykim-1: Will something like below work?
```
    current_acl = {k: v for k, v in (control_plane_acl or {}).items() if k in user_defined_keys}

    if configured_acl != current_acl:
        self.register_action(
            f"Updated LKE cluster {cluster.id} control plane ACL: {current_acl} -> {configured_acl}"
        )
        cluster.control_plane_acl_update(configured_acl)
```

ykim-1 (approved):
LGTM, tested both with and without customer tag

jriddle-linode (approved):
LGTM, works locally!

PR #529, opened 2024-06-25
Added note in documentation that Parent/Child support may not yet be generally available

ezilber-akamai:
## 📝 Description

Added support for optional module-level descriptions in info and list module base classes and updated docs for child_account_info and child_account_list


yec-akamai (approved):
Nice work! The base module update provides more flexibility.

zliang-akamai (approved):

PR #537, opened 2024-07-03
Inject documentation before release using `ansible-specdoc`

jriddle-linode:
## 📝 Description

Sets up injection step in release workflow to inject documentation into the module python files, required for ansible documentation and ansible_linode v1.0

**NOTE**: We're not entirely sure if this will render correctly in the ansible website, if it doesn't we'll have to disable this step and make an update to ansible-specdoc. Because of this I dont think the initial release including this PR should be v1.0.0 so we dont have to roll it back or jump straight to v1.0.1 like we have in the past.

Requires https://github.com/linode/ansible-specdoc/pull/38

## ✔️ How to Test

1. Use the make shortcut to inject each file
```bash
make deps
make inject
```
2. View a file in the `/plugins/modules/` folder and see the injected documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,5 +108,5 @@ endif
 
 inject:
 	@echo "Injecting documentation into source files"
-	for f in `ls ./plugins/modules/*.py`; do echo $$f; ansible-specdoc -j -i $$f; done
+	for f in `ls ./plugins/modules/*.py`; do ansible-specdoc -j -i $$f; done
 
--- a/Makefile
+++ b/Makefile
@@ -108,4 +108,5 @@ endif
 
 inject:
 	@echo "Injecting documentation into source files"
-	ansible-specdoc -j -i ./plugins/modules/*.py
+	for f in `ls ./plugins/modules/*.py`; do echo $$f; ansible-specdoc -j -i $$f; done
+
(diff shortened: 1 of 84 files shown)


Makefile
```diff
@@ -105,3 +105,8 @@ endif
 	@echo "api_url: $(TEST_API_URL)" >> $(INTEGRATION_CONFIG)
 	@echo "api_version: $(TEST_API_VERSION)" >> $(INTEGRATION_CONFIG)
 	@echo "ca_file: $(TEST_API_CA)" >> $(INTEGRATION_CONFIG)
+
+inject:
```
lgarber-akamai, line 109: Just to make sure things stay tidy, it make sense to also add a `inject-clean` that runs `ansible-specdoc -jc -i {FILE}` on each module to clear out the injected docs data.

plugins/module_utils/doc_fragments/ipv6_range_info.py
```diff
@@ -3,7 +3,7 @@
 specdoc_examples = ['''
 - name: Get info about an IPv6 range
   linode.cloud.ipv6_range_info:
-    range: 2600:3c01::''']
+    range: "2600:3c01::"''']
```
lgarber-akamai, line 6: It looks like this is causing the CI to fail — `make deps && make gendocs` should do the trick 🙂

jriddle-linode:
do not merge until https://github.com/linode/ansible-specdoc/pull/38 is merged and updated in ansible. Will cause release to fail until then.

lgarber-akamai:
@jriddle-linode Want us to do an ansible-specdoc release today to unblock this?

ykim-1 (approved):
LGTM, verified the new make commands work locally

lgarber-akamai (approved):
The `make inject`/`make inject-clean` targets seem to work well locally and all doc-related sanity tests are passing. Excellent work!

ezilber-akamai (approved):
Works well. Nice work!

PR #539, opened 2024-07-11
new: Support replication in `image` module

yec-akamai:
## 📝 Description

Allow replicate image by specifying a new field `regions_to_replicate`. User can wait for all replications to finish by specifying `wait_for_replications` and customize `wait_timeout`. 

## ✔️ How to Test

```
make TEST_ARGS="-v image_basic" test
```

Manual Test:
1. In a sandbox environment, i.e. dx-devenv, run the following playbook to create an image and replicate it.
```yaml
- name: test
  hosts: localhost
  tasks:
    - name: Create an instance to image
      linode.cloud.instance:
        label: my-test-ansible-instance
        region: us-east
        type: g6-standard-1
        image: linode/alpine3.19
        state: present
      register: instance_create

    - name: Create an image from the instance
      linode.cloud.image:
        label: my-gen2-image
        disk_id: '{{ instance_create.disks.0.id }}'
        description: 'cool'
        cloud_init: true
        tags:
          - 'image-gen2'
        state: present
      register: image_create

    - name: Replicate the image
      linode.cloud.image:
        label: my-gen2-image
        disk_id: '{{ instance_create.disks.0.id }}'
        description: 'cool2'
        tags:
          - 'test'
        replica_regions:
          - 'us-east'
          - 'eu-west'
        wait_for_replications: yes
        wait_timeout: 1200
        state: present
```
2. Observe that this image is replicate successfully.
3. Try to remove the available regions and put a new region
```yaml
    - name: Replicate the image
      linode.cloud.image:
        label: my-gen2-image
        disk_id: '{{ instance_create.disks.0.id }}'
        description: 'cool2'
        tags:
          - 'test'
        replica_regions:
          - 'us-central'
        wait_for_replications: yes
        wait_timeout: 1200
        state: present
```
4. Expect to see an error.
5. Try to remove one of the available region
```yaml
    - name: Replicate the image
      linode.cloud.image:
        label: my-gen2-image
        disk_id: '{{ instance_create.disks.0.id }}'
        description: 'cool2'
        tags:
          - 'test'
        replica_regions:
          - 'us-east'
        wait_for_replications: yes
        wait_timeout: 1200
        state: present
```
6. Observe that only one available region is returned.
7. Change around the script to discover any edge case.


plugins/modules/image.py
```diff
@@ -95,6 +95,17 @@
         editable=True,
         description=["A list of customized tags of this new Image."],
     ),
+    "regions_to_replicate": SpecField(
```
lgarber-akamai, line 98: nitpick: We could maybe name this field something like `replica_regions` or `replicate_to`?

yec-akamai: Good point! I want to keep the word `region` in the naming because it's what the endpoint calls it. Probably `replica_regions` works better in this case. I can also change this attribute name in TF to keep it consistent for user: https://github.com/linode/terraform-provider-linode/blob/proj/image-gen2/docs/resources/image.md?plain=1#L64

jriddle-linode: If we do that can we leave a comment with the actual field name? I know its obvious now but when i was going through the Docs for one of the PRs it did throw me off some i kept thinking I had the wrong endpoint.

lgarber-akamai (approved):
Looks perfect other than the above discussion around field naming, great work!

plugins/modules/image.py
yec-akamai: Thank you both for the suggestion! Just updated the name and added a comment.

lgarber-akamai (approved):
All tests and manual test steps (with a minor modification to remove `{{ r }}` are passing on my end, great work!

ykim-1 (approved):
LGTM, manual and tests work locally. Nice work!

PR #550, opened 2024-07-24
Implemented changes for Linode Disk Encryption

ezilber-akamai:
## 📝 Description

Updated ansible modules to consume changes according to the API Spec (can be found in the ticket).

### Integration Tests
- `make TEST_ARGS="-v instance_disk_encryption" test`
- `make TEST_ARGS="-v lke_cluster_basic" test`
- `make TEST_ARGS="-v lke_node_pool_basic" test`


plugins/modules/instance.py
```diff
@@ -83,6 +83,12 @@
         type=FieldType.string,
         description=["The filesystem to create this disk with."],
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        description="The disk encryption status of this disk."
+        + "NOTE: Disk encryption may not currently be available to all users.",
+        choices=["enabled", "disabled", "pde"],
+    ),
```
yec-akamai, line 91: In the sepc doc this field seems only accept two values: `disabled`,`enabled`. And we probably can set its default to be `enabled`.

plugins/modules/instance.py
```diff
@@ -523,6 +529,16 @@
         suboptions=linode_instance_placement_group_spec,
         description=["A Placement Group to create this Linode under."],
     ),
+    "lke_cluster_id": SpecField(
+        type=FieldType.integer,
+        description="The id of the LKE cluster associated with this Linode, if there is one.",
+    ),
```
yec-akamai, line 535: `lke_cluster_id` seems a read-only field here. We probably don't need to put it as a SpecField.

plugins/modules/instance.py
```diff
@@ -523,6 +529,16 @@
...
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        description="The disk encryption status of this Linode. "
+        + "NOTE: Disk encryption may not currently be available to all users.",
+        choices=["enabled", "disabled", "pde"],
+    ),
```
yec-akamai, line 541: Looks like a redundant one

plugins/modules/lke_cluster.py
```diff
@@ -133,6 +133,12 @@
         ],
         suboptions=linode_lke_cluster_autoscaler,
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        description="The disk encryption status of this node pool. "
+        + "NOTE: Disk encryption may not currently be available to all users.",
+        choices=["enabled", "disabled", "pde"],
+    ),
```
yec-akamai, line 141: `disk_encryption` seems a read-only field for lke_cluster. We don't have to put it as a SpecField here.

plugins/modules/lke_node_pool.py
```diff
@@ -140,6 +140,12 @@
         ],
         default=600,
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        description="The disk encryption status of this node pool. "
+        + "NOTE: Disk encryption may not currently be available to all users.",
+        choices=["enabled", "disabled", "pde"],
+    ),
```
yec-akamai, line 148: Also read-only. We should be good to remove it here too.

yec-akamai (commented):
Probably need to change a bit the SpecFields based on the API spec doc:

plugins/modules/instance.py
ezilber-akamai: The API spec mentioned `pde` as one of the possible values in the "Schema" section. Initially I had it set a default value but this was causing some problems because the API would recognize creating an instance without specifying the `disk_encryption` status as an update.

plugins/modules/instance.py
ezilber-akamai: It looks like disk_encryption should be able to be set either directly in the linode instance or within a disk that is attached to it. In order to allow this functionality, I had to add it to both the instance and instance_disk specs.

plugins/modules/lke_cluster.py
ezilber-akamai: Interesting. I updated some tests to create lke_clusters with disk_encryption set and they seemed to be working. Where is it listed that `disk_encryption` is read-only for `lke_cluster` and `lke_node_pool`?

yec-akamai:
I can't share the spec doc on github for security reason. Let's take the discussion offline!

yec-akamai (approved):
Looks great! Tests all passed locally 🎉

plugins/modules/instance.py
```diff
@@ -83,6 +83,13 @@
         type=FieldType.string,
         description=["The filesystem to create this disk with."],
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        description="The disk encryption status of this disk."
+        + "NOTE: Disk encryption may not currently be available to all users.",
+        choices=["enabled", "disabled"],
+        default="enabled",
```
lgarber-akamai, line 91: It might make sense to defer the default logic to the API since there could be a case where a user wants to update certain fields of an LDE-disabled instance without implicitly trying to update `disk_encryption`, which raises an error 👀

ezilber-akamai: Good call! Fixed.

lgarber-akamai (approved):
Tests are passing locally and I haven't been able to break it in my manual testing. Excellent work!

PR #552, opened 2024-07-29
fix: Fix grants AttributeError; migrate user_info to base InfoModule;

lgarber-akamai:
## 📝 Description

This pull request resolves the following error that was being raised when attempting to use the `linode.cloud.user_info` module:

```
AttributeError: 'PaginatedList' object has no attribute 'grants'
```

Because we had already planned to migrate user_info to the `InfoModule` base (which replaces the old result gathering logic) I decided to quickly do it as a part of this PR. If anyone would rather me handle it in a separate PR definitely let me know 🙂 

## ✔️ How to Test

### Integration Tests

```
make TEST_ARGS="-v user_info" test
```

### Manual Testing

1. In an ansible_linode sandbox environment (e.g. dx-devenv), run the following playbook:

```yaml
  hosts: localhost
  tasks:
    - set_fact:
        id: "{{ 1000000000 | random }}"

    - name: Create Linode User
      linode.cloud.user:
        username: 'test-user-{{ id }}'
        email: 'test-user-{{ id }}@linode.com'
        state: present
      register: user_create

    - name: Get info about the user
      linode.cloud.user_info:
        username: '{{ user_create.user.username }}'
      register: user_info

    - name: Remove the user
      linode.cloud.user:
        username: '{{ user_create.user.username }}'
        state: absent

    - name: Pretty-print the user's info
      debug:
        var: user_info
```

2. Ensure the playbook succeeds.
3. Ensure the `user` and `grants` is shown in the output of the last task.

### DIFF
--- a/plugins/modules/user_info.py
+++ b/plugins/modules/user_info.py
@@ -5,56 +5,52 @@
 
 from __future__ import absolute_import, division, print_function
 
-from typing import Any, List, Optional
-
 import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.user_info as docs
-from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
-    LinodeModuleBase,
+from ansible_collections.linode.cloud.plugins.module_utils.linode_common_info import (
+    InfoModule,
+    InfoModuleAttr,
+    InfoModuleResult,
 )
-from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import (
-    global_authors,
-    global_requirements,
-)
-from ansible_specdoc.objects import (
-    FieldType,
-    SpecDocMeta,
-    SpecField,
-    SpecReturnValue,
-)
+from ansible_specdoc.objects import FieldType
 from linode_api4 import User
 
-spec = {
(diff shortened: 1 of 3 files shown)


plugins/modules/user_info.py
```diff
@@ -5,97 +5,59 @@
...
             docs_url="https://techdocs.akamai.com/linode-api/reference/get-user-grants",
-            type=FieldType.dict,
-            sample=docs.result_grants_samples,
-        ),
-    },
+            samples=docs.result_grants_samples,
+            get=lambda client, user, params: client.get(
+                # We can't use the UserGrants type here because
+                # it does not serialize directly to JSON or store
+                # the API response JSON.
```
zliang-akamai, line 35: Do we want to make UserGrants serializable as a Python SDK issue?

lgarber-akamai: Good point! Yeah I think that's a good idea

zliang-akamai (approved):
LGTM, tested and it works well!

yec-akamai (approved):
Works nicely on my end!

PR #558, opened 2024-08-09
ref: Migrate `profile_info` to use common info module

yec-akamai:
## 📝 Description

Migrate `profile_info` to use common info module. Also make some necessary change to the base info module because profile_info doesn't require any attribute or parameter to call. 

## ✔️ How to Test

```make TEST_ARGS="-v profile_info" test ```

Manual test:
1. In a sandbox environment, run the following script to get the info of current profile:
```yaml
- name: test
  hosts: localhost
  tasks:
    - name: Get profile_info about current profile
      linode.cloud.profile_info:
      register: profile
```
2. Observe that profile info the successfully returned.

### DIFF
--- a/plugins/module_utils/linode_common_info.py
+++ b/plugins/module_utils/linode_common_info.py
@@ -72,7 +72,8 @@ class InfoModuleResult:
         samples (Optional[List[str]]): A list of sample results for this field.
         get (Optional[Callable]): A function to call out to the API and return the data
                                   for this field.
-                                  NOTE: This is only relevant for secondary results.
+                                  NOTE: This is only relevant for secondary results
+                                  or primary result without any attributes.
     """
 
     field_name: str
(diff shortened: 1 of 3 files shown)


plugins/module_utils/linode_common_info.py
```diff
@@ -220,10 +231,14 @@ def run(self) -> None:
...
+            if len(self.attributes) > 0
+            else None
+        )
 
         super().__init__(
             module_arg_spec=self.module_arg_spec,
-            required_one_of=[attribute_names],
-            mutually_exclusive=[attribute_names],
+            required_one_of=attribute_names,
+            mutually_exclusive=attribute_names,
```
yec-akamai, line 243: The previous implementation brought an edge case that when the module didn't require any attribute, we still send an `[[]]` value as `required_one_of`. It caused a bug to fail the module.

ezilber-akamai (approved):
Integration and manual tests working properly. Nice work!

jriddle-linode (approved):
LGTM works locally.

PR #562, opened 2024-08-14
Migrate `account_info` to use common info module

ezilber-akamai:
## 📝 Description

Migrated `account_info` module to use the common info module.

### Integration Test
`make TEST_ARGS="-v account_info" test`

### Manual Test

1. In an ansible_linode sandbox environment (e.g. dx-devenv), run the following:
```
- name: test
  hosts: localhost
  tasks:
    - name: Get info about current account
      linode.cloud.account_info:
      register: account_info
```
2. Ensure that the account info is successfully returned.

### DIFF
--- a/docs/modules/account_info.md
+++ b/docs/modules/account_info.md
@@ -22,7 +22,7 @@ Get info about a Linode Account.
 
 ## Return Values
 
-- `account` - The account info in JSON serialized form.
+- `account` - The returned Account.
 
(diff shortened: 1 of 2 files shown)


lgarber-akamai (approved):
Test passed after removing the skip and re-adding the environment section, nice work!

yec-akamai (approved):
Nice work!

PR #575, opened 2024-08-22
ref: Migrate object_cluster_list module to ListModule base class; add deprecation support to ListModule and InfoModule

lgarber-akamai:
## 📝 Description

This pull request migrates the `object_cluster_list` module over to the ListModule base class. Because this module is deprecated, this PR also adds support for deprecating modules using the ListModule and InfoModule base classes.

**NOTE: object_cluster_info was intentionally not migrated as a part of this PR because its usage deviates too much from InfoModule.**

## ✔️ How to Test

### Integration Testing

1. Open `tests/integration/targets/object_cluster_list/main.yaml`, remove the debug task and uncomment the rest of the playbook.
2. Run the tests:

```
make TEST_ARGS="-v object_basic object_cluster_list" test
```

### DIFF
--- a/tests/integration/targets/object_cluster_list/tasks/main.yaml
+++ b/tests/integration/targets/object_cluster_list/tasks/main.yaml
@@ -10,18 +10,6 @@
 #    - assert:
 #        that:
 #          - no_filter.clusters | length >= 1
-#
-#    - name: List regions with filter on region
-#      linode.cloud.object_cluster_list:
-#        filters:
-#          - name: region
-#            values: us-ord
-#      register: filter
-#
-#    - assert:
-#        that:
-#          - filter.clusters | length >= 1
-#          - filter.clusters[0].region == 'us-ord'
 
   environment:
     LINODE_UA_PREFIX: '{{ ua_prefix }}'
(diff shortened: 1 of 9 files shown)


tests/integration/targets/object_cluster_list/tasks/main.yaml
```diff
@@ -10,18 +10,6 @@
 #    - assert:
 #        that:
 #          - no_filter.clusters | length >= 1
-#
-#    - name: List regions with filter on region
```
lgarber-akamai, line 14: This no longer functions in both the old and new module implementations. Given this is a deprecated API, do we think it's worth keeping and investigating further?

yec-akamai: I think we can skip and save effort since it's deprecated.

yec-akamai (approved):
Works as expected!

tests/integration/targets/object_cluster_list/tasks/main.yaml
lgarber-akamai: Sounds good to me!

jriddle-linode (approved):
Looks much better and works locally.

PR #584, opened 2024-09-03
ref: Remove state,tags,label from linode_common

yec-akamai:
## 📝 Description

Because of the variety of our modules, we don't require state, tags and label in each of them. Remove these three fields from the linode_common and only implement in those which required. It helps us get rid of some ugly workaround and improves readability. 

## ✔️ How to Test

Tests the modules which are added the state field into the module spec:

```
make TEST_ARGS="-v lke_cluster_basic" test
make TEST_ARGS="-v placement_group_basic" test
make TEST_ARGS="-v placement_group_assign" test
```

### DIFF
--- a/plugins/modules/ip_share.py
+++ b/plugins/modules/ip_share.py
@@ -25,9 +25,6 @@ from ansible_specdoc.objects import (
 from linode_api4.objects import Instance
 
 ip_share_spec = {
-    # Disable the default values
-    "label": SpecField(type=FieldType.string, required=False, doc_hide=True),
-    "state": SpecField(type=FieldType.string, required=False, doc_hide=True),
     "ips": SpecField(
         type=FieldType.list,
         required=True,
--- a/plugins/modules/user.py
+++ b/plugins/modules/user.py
@@ -165,11 +165,6 @@ SPEC_GRANTS = {
 }
 
 SPEC = {
-    # We don't use label for this module
-    "label": SpecField(
-        type=FieldType.string,
-        doc_hide=True,
-    ),
     "username": SpecField(
         type=FieldType.string,
         required=True,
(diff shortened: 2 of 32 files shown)


lgarber-akamai:
Integration suite run: https://github.com/linode/ansible_linode/actions/runs/10690025027

plugins/modules/ip_share.py
```diff
@@ -25,9 +25,6 @@
 from linode_api4.objects import Instance
 
 ip_share_spec = {
-    # Disable the default values
-    "label": SpecField(type=FieldType.string, required=False, doc_hide=True),
-    "state": SpecField(type=FieldType.string, required=False, doc_hide=True),
```
lgarber-akamai, line 30: It looks like state is actually used in this module contrary to this snippet, so it might make sense to bring it back as a normal field 👍

yec-akamai: Nice catch! Just fixed it and verified locally

plugins/modules/user.py
```diff
@@ -165,11 +165,6 @@
 }
 
 SPEC = {
```
lgarber-akamai, line 167: I'm seeing an error on the user_basic test now:

```
TASK [user_basic : Create Linode User] *****************************************
fatal: [testhost]: FAILED! => {"changed": false, "msg": "failed to create user: AccountGroup.user_create() got an unexpected keyword argument 'state'"}
```

I'm assuming the state just needs to be filtered from the creation kwargs 👍

lgarber-akamai (approved):
All tests are passing on my end. LGTM after the linting failure is addressed!

ezilber-akamai (approved):
Everything is passing locally. Nice work!

PR #589, opened 2024-09-24
Refactor image replication test cases

yec-akamai:
## 📝 Description

Remove the hardcoded region and instead using core region with Object Storage capability for image replication. 

## ✔️ How to Test

```
make TEST_ARGS="-vvv image_basic" test
```

### DIFF
--- a/plugins/module_utils/linode_common.py
+++ b/plugins/module_utils/linode_common.py
@@ -104,7 +104,7 @@ RETRY_INTERVAL_SECONDS = float(4)
 RETRY_STATUSES = {408, 429, 502}
 
 
-class LinodeModuleBase:  # pylint: disable=too-many-positional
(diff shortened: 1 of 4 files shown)


plugins/module_utils/linode_common.py
```diff
@@ -104,7 +104,7 @@
 RETRY_STATUSES = {408, 429, 502}
 
 
-class LinodeModuleBase:
```
zliang-akamai, line 107: Did pylint complain about this?

zliang-akamai (approved):
LGTM!

plugins/module_utils/linode_common.py
ykim-1: I also included the lint fix in Makefile (#592)

ykim-1 (approved):
LGTM, verified locally. Thanks for fixing the tests!

PR #605, opened 2024-10-08
Configure max-positional-arguments in `pyproject.toml`

zliang-akamai:
## 📝 Description

This is to centralize the configuration of pylint into `pyproject.toml` and cleanup the argument passing from `Makefile` that configures the similar thing.

## ✔️ How to Test

```bash
make lint
```


lgarber-akamai (approved):
Always nice to tidy things up, nice work!

ykim-1 (approved):

PR #609, opened 2024-10-24
Fix an example in the doc of list regions module

zliang-akamai:
## 📝 Description

This is to fix an example in the documentation of the list regions module.

## ✔️ How to Test

Run `ansible-playbook /path/to/main.yaml -vvv` against the YAML file with the following content.

```yaml
---
  hosts: localhost
  gather_facts: no
  tasks:
    - name: Resolve all Linode regions
      linode.cloud.region_list:
        filters:
          - name: site_type
            values: core
```


ykim-1 (approved):
LGTM

jriddle-linode (approved):

PR #619, opened 2024-11-05
Cleanup, upgrade, and re-organize dependencies

zliang-akamai:
## 📝 Description
This is to organize all dependencies

integration test: https://github.com/linode/ansible_linode/actions/runs/11690045146

1. Downgrade `types-requests` to `2.32.0.20240914` because newer version requiring urllib3>=2, which is not compatible with Python 3.9.
2. Move `types-requests` to dev requirements becauase it's only needed for linters.
3. Upgrade `py-version` to `3.9` for `pylint`.
4. Remove EOL Python versions and add `3.12` for `black`.
5. Pin version of `polling` to 0.3.2 because it's a third party package. (we should consider its fork, [`polling2`](https://github.com/ddmee/polling2), because `polling` is no longer maintained)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,11 +5,10 @@ line_length = 80
 [tool.black]
 line-length = 80
 target-version = [
-    "py37",
-    "py38",
     "py39",
     "py310",
     "py311",
+    "py312",
 ]
 exclude = "plugins/module_utils/doc_fragments"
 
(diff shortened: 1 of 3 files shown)


lgarber-akamai (approved):
Looks great, thanks for addressing this!

ykim-akamai (approved):
LGTM, thanks for looking into this!

PR #625, opened 2024-11-21
Pin Python version to 3.10 in documentation validation workflow

lgarber-akamai:
## 📝 Description

**NOTE: This changes from the PR will _not_ apply to workflows on this PR until it has been merged to main.**

This pull request pins the Python version in the documentation validation workflow to `3.10`, which matches the default version defined in our [integration test workflow](https://github.com/linode/ansible_linode/blob/dev/.github/workflows/integration-tests.yml#L23). 

This is necessary because [Ansible 11.0.0](https://pypi.org/project/ansible/11.0.0/) does not support < Python 3.11 but has some [minor documentation differences](https://github.com/linode/ansible_linode/actions/runs/11956754987/job/33332255116#step:10:13) when compared against 10.x.x, causing our documentation validation workflows to fail.

In the future we should consider bumping the minimum Python version of this repo to match Ansible's minimum Python version.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -196,59 +196,30 @@ jobs:
       - name: Notify Slack
         uses: slackapi/slack-github-action@v2.0.0
         with:
-          channel-id: ${{ secrets.SLACK_CHANNEL_ID }}
+          method: chat.postMessage
+          token: ${{ secrets.SLACK_BOT_TOKEN }}
           payload: |
-            {
-              "blocks": [
-                {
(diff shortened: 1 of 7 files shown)


zliang-akamai (approved):
LGTM!

zliang-akamai:
BTW let's backlog a ticket for upgrading Python and Ansible?

If Ansible 11 has a different docs format, I wonder if it's backward compatible with Ansible < 11

ykim-akamai (approved):
LGTM, thanks for looking into this

PR #626, opened 2024-11-21
doc: Remove Image Services Gen2 LA notice

yec-akamai:
## 📝 Description

Image Services Gen2 goes generally available 🎉


ezilber-akamai (approved):
The docs validation issue should be fixed by #625, looks good otherwise!

jriddle-linode (approved):

PR #628, opened 2024-12-05
new: Support IP module to allocate a new IP

yec-akamai:
## 📝 Description

Support a new linode.cloud.ip module to allocate a new IP to a Linode instance.

## ✔️ How to Test

Integration test:
```
make TEST_ARGS="ip_basic" test
```

Manual test:

1. In a sandbox environment, i.e. dx-dvenv, run the following script to create a Linode instance and allocate a new IP to it.
```yaml
    - name: Create a Linode Instance
      linode.cloud.instance:
        label: 'ansible-instance-test'
        region: us-southeast
        type: g6-standard-1
        image: linode/alpine3.19
        state: present
      register: create_instance

    - name: Allocate a new IP to the Linode
      linode.cloud.ip:
        linode_id: '{{ create_instance.instance.id }}'
        public: true
        type: ipv4
        state: present
      register: allocate_ip
```

2. Observe the ip is allocated successfully.
3. Clean up the resources created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,9 @@ test: integration-test
 testall:
 	./scripts/test_all.sh
 
+smoketest:
+	./scripts/test_basic_smoke.sh
+
 unittest:
 	ansible-test units --target-python default
 
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ delete-e2e-firewall: update-test-submodules
 		echo "$$OUTPUT"; \
 		exit 1; \
 	else \
-		echo "E2E Cloud firewall created successfully."; \
+		echo "E2E Cloud firewall deleted successfully."; \
 	fi
 
 update-test-submodules:
--- a/docs/inventory/instance.rst
+++ b/docs/inventory/instance.rst
@@ -76,19 +76,19 @@ Parameters
 
 
       **parent_group (type=str):**
-        \• parent group for keyed group
+        \• parent group for keyed group.
 
 
       **prefix (type=str):**
-        \• A keyed group name will start with this prefix
+        \• A keyed group name will start with this prefix.
 
 
       **separator (type=str, default=_):**
-        \• separator used to build the keyed group name
+        \• separator used to build the keyed group name.
 
 
       **key (type=str):**
-        \• The key from input dictionary used to generate groups
+        \• The key from input dictionary used to generate groups.
 
 
       **default_value (type=str):**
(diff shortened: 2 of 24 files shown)


Makefile
```diff
@@ -91,7 +91,7 @@ delete-e2e-firewall: update-test-submodules
 		echo "$$OUTPUT"; \
 		exit 1; \
 	else \
-		echo "E2E Cloud firewall created successfully."; \
+		echo "E2E Cloud firewall deleted successfully."; \
```
yec-akamai, line 94: nitpick

docs/inventory/instance.rst
```diff
@@ -58,7 +58,7 @@ Parameters
 
 
   **strict (type=bool):**
```
lgarber-akamai, line 60: @yec-akamai When you get a chance could this file's diff be removed from this PR to fix the docs validation workflow?

Just for context, this is a result of the latest version of Ansible not supporting our minimum Python version but having some minor documentation differences. I think we're planning on bumping our minimum Python version in the future 👍 

See #625 for a bit more context

plugins/modules/ip.py
```diff
@@ -0,0 +1,107 @@
...
+    ),
+    "type": SpecField(
+        type=FieldType.string,
+        required=True,
+        choices=["ipv4"],
+        description=[
+            "The type of address you are requesting. "
+            "Only IPv4 addresses may be allocated through this operation."
+        ],
+    ),
```
lgarber-akamai, line 44: Could a `state` attribute be added here to support creating/deleting IPs?

yec-akamai: I find delete IP is not properly implemented in python sdk. We can't use the delete() function from Base because the endpoint is not correct. I'll create a python-sdk ticket to implement the right endpoint.

plugins/modules/ip.py
```diff
@@ -0,0 +1,134 @@
...
+            self.register_action(
+                f"IP allocation to Linode {linode_id} completed."
+            )
+        except Exception as exc:
+            self.fail(msg=f"failed to allocate IP to Linode {linode_id}: {exc}")
+
+        self.results["ip"] = ip._raw_json
+
+    def _handle_absent(self) -> None:
+        # TODO: Implement deleting IP once it's available in python-sdk
```
lgarber-akamai, line 111: Could we raise an error here until the issue in the Python SDK has been resolved?

lgarber-akamai (approved):
Looks great and tests are passing on my end!

ykim-akamai (approved):
LGTM, tests passing locally

PR #645, opened 2025-02-10
Implement linode.cloud.database_mysql_v2 module

lgarber-akamai:
## 📝 Description

This pull request implements a new `linode.cloud.database_mysql_v2` module supporting the new Aiven-backed managed databases service.

## ✔️ How to Test

### Integration Testing

```
make TEST_ARGS="-v database_mysql_v2_basic database_mysql_v2_complex" test-int
```

### DIFF
--- a/plugins/modules/database_mysql_v2.py
+++ b/plugins/modules/database_mysql_v2.py
@@ -188,10 +188,6 @@ class Module(LinodeModuleBase):
             "database", "database_create"
         )
 
-        import q
-
-        q.q(params)
-
         database = self.client.database.mysql_create(**params)
 
         create_poller.set_entity_id(database.id)
--- a/plugins/modules/database_mysql_v2.py
+++ b/plugins/modules/database_mysql_v2.py
@@ -1,0 +1,330 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""This module contains the implementation for the linode.cloud.database_mysql_v2 module."""
+
+from __future__ import absolute_import, division, print_function
+
+import copy
+from typing import Any, Optional
+
+import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.database_mysql_v2 as docs
+from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
+    LinodeModuleBase,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_database_shared import (
+    SPEC_FORK,
+    SPEC_UPDATE_WINDOW_V2,
+    call_protected_provisioning,
+)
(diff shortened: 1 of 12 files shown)


plugins/modules/database_mysql_v2.py
```diff
@@ -0,0 +1,330 @@
...
+            "The Linode Instance type used by the Managed Database for its nodes."
+        ],
+        editable=True,
+    ),
+    "fork": SpecField(
+        type=FieldType.dict,
+        description=["Information about a database to fork from."],
+        suboptions=SPEC_FORK,
+    ),
+    "updates": SpecField(
```
lgarber-akamai, line 86: I was debating adding a field that allows users to disable polling, but I figured it might be more effort than it's worth given DB provisioning times are much faster now and it would introduce a bunch of update timing issues.

plugins/modules/database_mysql_v2.py
```diff
@@ -0,0 +1,330 @@
...
+        database: MySQLDatabase,
+        desired_status: str,
+        timeout: int = 30 * 60,
+        step: int = 2,
+    ) -> None:
+        def __poll_status() -> bool:
+            database._api_get()
+            return database.status == desired_status
+
+        poll_condition(__poll_status, timeout=timeout, step=step)
```
lgarber-akamai, line 167: We might be able to reuse this in the PostgreSQL module if we use a TypeVar for the database type.

ezilber-akamai: Cool! I will use this once this PR is merged.

plugins/modules/database_mysql_v2.py
ezilber-akamai: Yeah I agree, I think its not worth doing.

lgarber-akamai:
Converting back to a draft while we investigate some weird resize issues

tests/integration/targets/database_mysql_v2_complex/tasks/main.yaml
```diff
@@ -0,0 +1,188 @@
...
+        fork:
+          source: "{{ db_refresh.database.id }}"
+        state: present
+      register: db_forked
+
+    - name: Assert a new forked database has been created
+      assert:
+        that:
+          - db_forked.changed == True
+          - db_forked.database.id == db_refresh.database.id
```
ezilber-akamai, line 167: I don't think this assertion should be here. The ids of the two databases should be different.

ezilber-akamai (approved):
Tests are passing locally. Nice work!

yec-akamai (approved):
Tested and works well on my end. Nice work!

PR #657, opened 2025-03-18
Add support for LKE-E features

ezilber-akamai:
## 📝 Description

Added support for listing LKE versions with `tier` specified. Added support for getting LKE version info with and without tier specified. Added support for specifying tier upon LKE cluster creation.

Also changed the `high_availability` field to no longer default to false since `enterprise` clusters have `high_availability` set to true by default.

## ✔️ How to Test
The following steps assume you have this PR pulled down locally.

### Integration Testing
`make test-int TEST_SUITE="lke_version_info"`
`make test-int TEST_SUITE="lke_version_list"`
`make test-int TEST_SUITE="lke_cluster_enterprise"` **_NOTE:_** This test will a while to run.
`make test-int TEST_SUITE="lke_node_pool_enterprise"` **_NOTE:_** This test will a while to run.

### Manual Testing
1. In a local testing environment (i.e. dx-devenv sandbox), run the following ansible playbook:
```
    # LKE Version Get tests
    - name: Get an LKE Version without specifying tier
      linode.cloud.lke_version_info:
        id: '1.31'
      register: lke_version_no_tier
      
    - name: Get a standard LKE Version
      linode.cloud.lke_version_info:
        api_version: v4beta
        id: '1.31'
        tier: 'standard'
      register: lke_version_tier_standard
      
    - name: Get an enterprise LKE Version
      linode.cloud.lke_version_info:
        api_version: v4beta
        id: 'v1.31.1+lke4'
        tier: 'enterprise'
      register: lke_version_tier_enterprise


    # LKE Versions List tests
    - name: List all LKE Versions
      linode.cloud.lke_version_list:
      register: all_lke_versions

    - name: List all LKE standard Versions
      linode.cloud.lke_version_list:
        api_version: v4beta
        tier: "standard"
      register: lke_standard_versions

    - name: List all LKE enterprise Versions
      linode.cloud.lke_version_list:
        api_version: v4beta
        tier: "enterprise"
      register: lke_enterprise_versions


    # Create an LKE Enterprise cluster
    - name: Create a Linode LKE Enterprise cluster
      linode.cloud.lke_cluster:
        api_version: v4beta
        tier: 'enterprise'
        label: 'ansible-test-enterprise'
        region: 'us-ord'
        k8s_version: 'v1.31.1+lke4'
        node_pools:
          - type: g6-standard-1
            count: 3
          - type: g6-standard-4
            count: 1
        skip_polling: true
        state: present
      register: create_enterprise_cluster


    # Create an LKE Enterprise node pool
    - name: Add an enterprise node pool to the cluster
    linode.cloud.lke_node_pool:
      cluster_id: '{{ create_enterprise_cluster.cluster.id }}'
      tags: [ 'my-pool' ]
      type: g6-standard-1
      count: 2
      k8s_version: '{{ lke_version_tier_enterprise }}'
      update_strategy: 'rolling_update'
      skip_polling: false
      state: present
    register: new_pool    
```
2. Verify the output for the `info` and `list` modules looks correct.
3. Verify that the LKE Enterprise cluster was successfully created in Cloud Manager.
4. Verify that the LKE Enterprise node pool was successfully created in Cloud Manager.
5. Clean up any resources created for this test.

### DIFF
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -258,6 +258,7 @@ linode_lke_cluster_spec = {
             "and can only be used with v4beta.",
         ],
         editable=False,
+        required=False,
         choices=["standard", "enterprise"],
         default="standard",
     ),
@@ -311,6 +312,7 @@ CREATE_FIELDS: Set[str] = {
     "control_plane",
     "high_availability",
     "apl_enabled",
+    "tier"
 }
 
 DOCUMENTATION = r"""
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -259,7 +259,6 @@ linode_lke_cluster_spec = {
         editable=False,
         required=False,
         choices=["standard", "enterprise"],
-        default="standard",
     ),
 }
 
@@ -382,7 +381,8 @@ class LinodeLKECluster(LinodeModuleBase):
 
         label = params.pop("label")
 
-        # Determine high_availability based on tier
+        # We must determine the default value of high_availability here because
+        # the default value is False if the tier is standard, and True if the tier is enterprise
         tier = params.get("tier")
         high_availability = tier not in (
             "standard",
@@ -505,17 +505,8 @@ class LinodeLKECluster(LinodeModuleBase):
 
         pools = new_params.pop("node_pools")
 
-        # These are handled separately
+        # This is handled separately
         new_params.pop("k8s_version")
-
-        # Determine high_availability from tier
-        tier = self.module.params.get("tier")
-        high_avail = tier != "standard"
-
-        if high_avail is not None:
-            new_params["high_availability"] = (
-                high_avail  # Only include if explicitly set
-            )
 
         handle_updates(
             cluster, new_params, MUTABLE_FIELDS, self.register_action
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -251,6 +251,13 @@ linode_lke_cluster_spec = {
         ],
         default=False,
     ),
+    "tier": SpecField(
+        type=FieldType.string,
+        description=["The desired tier of the LKE Cluster."],
+        editable=False,
+        choices=["standard", "enterprise"],
+        default="standard",
+    ),
 }
 
 SPECDOC_META = SpecDocMeta(
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -205,7 +205,6 @@ linode_lke_cluster_spec = {
             "Defines whether High Availability is enabled for the "
             "Control Plane Components of the cluster. "
         ],
-        default=False,
     ),
     "acl": SpecField(
         type=FieldType.dict,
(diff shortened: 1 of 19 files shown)


plugins/modules/lke_cluster.py
```diff
@@ -487,7 +507,15 @@ def _update_cluster(self, cluster: LKECluster) -> None:
...
-        new_params.pop("high_availability")
+
+        # Determine high_availability from tier
+        tier = self.module.params.get("tier")
+        high_avail = tier != "standard"
+
+        if high_avail is not None:
+            new_params["high_availability"] = (
+                high_avail  # Only include if explicitly set
+            )
```
yec-akamai, line 518: Is this logic necessary to have here? It looks like `ha` is handled by `_cluster_put_updates ` not the `handle_updates`, and it's not one of the MUTABLE_FIELDS.

yec-akamai (commented):
Thank you for handling the special cases! Just a few comments:

plugins/modules/lke_cluster.py
```diff
@@ -251,6 +249,18 @@
...
+        type=FieldType.string,
+        description=[
+            "The desired tier of the LKE Cluster.",
+            "NOTE: LKE Enterprise may not currently be available to all users ",
+            "and can only be used with v4beta.",
+        ],
+        editable=False,
+        required=False,
+        choices=["standard", "enterprise"],
+        default="standard",
```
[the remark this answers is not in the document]
ezilber-akamai: Good point, I will remove it.

plugins/modules/lke_cluster.py
```diff
@@ -371,9 +382,18 @@ def _create_cluster(self) -> Optional[LKECluster]:
 
         label = params.pop("label")
 
+        # Determine high_availability based on tier
+        tier = params.get("tier")
+        high_availability = tier not in (
+            "standard",
+            None,
+        )  # False if "standard", True otherwise
```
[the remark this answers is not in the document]
ezilber-akamai: Sounds good! I'll add some detail to the comment.

plugins/modules/lke_cluster.py
ezilber-akamai: I think you're right, we don't need the logic here. Good catch!

tests/integration/targets/lke_version_info/tasks/main.yaml
```diff
@@ -0,0 +1,44 @@
+- name: lke_version_info
```
yec-akamai, line 1: I wonder if it's safer to call the list module first then test the get module. Just in case the valid k8s version change and break our test in the future 😢

ezilber-akamai: Good point. Fixed!

tests/integration/targets/lke_node_pool_basic/tasks/main.yaml
```diff
@@ -54,6 +54,10 @@
         state: present
       register: new_pool
 
+    - name: Output the nodes in the new pool
+      debug:
+        var: new_pool.node_pool.nodes
+
```
yec-akamai, line 60: shall we clean this up? It looks like for debugging

yec-akamai (commented):
I'm currently getting `[500] Internal server error` with nodepools endpoints 😞 The implementation looks good to me though.

tests/integration/targets/lke_node_pool_basic/tasks/main.yaml
ezilber-akamai: Oops. Yeah I will remove this

ezilber-akamai:
> I'm currently getting `[500] Internal server error` with nodepools endpoints 😞 The implementation looks good to me though.

Interesting. It was working when I reran the tests yesterday. I will try running the test again.

yec-akamai (approved):
Great work! Tests now passing locally 🎉

ykim-akamai (approved):
LGTM, tests verified locally. Nice work!

PR #659, opened 2025-04-01
Add DBaaS Suspend and Resume

jriddle-linode:
## 📝 Description

Allows suspending and resuming a database via the status field in ansible

## ✔️ How to Test

```bash
make test-int TEST_SUITE="database_postgresql_v2_basic"
```
```bash
make test-int TEST_SUITE="database_mysql_v2_basic"
```

### DIFF
--- a/plugins/modules/database_mysql_v2.py
+++ b/plugins/modules/database_mysql_v2.py
@@ -40,7 +40,7 @@ from linode_api4 import MySQLDatabase
 SPEC = {
     "state": SpecField(
         type=FieldType.string,
-        choices=["present", "absent"],
+        choices=["resume", "suspend", "present", "absent"],
         required=True,
         description=["The desired state of the Managed Database."],
     ),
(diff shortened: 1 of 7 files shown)


plugins/modules/database_mysql_v2.py
```diff
@@ -326,12 +326,46 @@ def _handle_absent(self) -> None:
...
         state = kwargs.get("state")
 
         if state == "absent":
             self._handle_absent()
+        if state == "suspend":
+            self._handle_suspend()
+        if state == "resume":
+            self._handle_resume()
         else:
             self._handle_present()
```
zliang-akamai, line 370:
```suggestion
        elif state == "suspend":
            self._handle_suspend()
        elif state == "resume":
            self._handle_resume()
        else:
            self._handle_present()
```

Otherwise `_handle_present` will always be called when state is not `"resume"`..

plugins/modules/database_mysql_v2.py
```diff
@@ -326,12 +326,46 @@ def _handle_absent(self) -> None:
...
+            self.client.database.mysql_instances,
+            MySQLDatabase.label == params.get("label"),
+        )
+
+        if database is not None:
+            self._populate_results(database)
+
+            database.resume()
+
+            self.register_action(f"Resumed MySQL database {database.id}")
```
zliang-akamai, line 357: Optional thing for consideration, is it worth to combine some parts of these functions?

```suggestion
    def _handle_suspend_or_resume(self, state: str) -> None:
        params = self.module.params

        database = safe_find(
            self.client.database.mysql_instances,
            MySQLDatabase.label == params.get("label"),
        )

        if database is not None:
            self._populate_results(database)
            
            if state == "suspend":
                _handle_suspend(self, database)
            elif state == "resume":
                _handle_resume(self, database)

    def _handle_suspend(self, database) -> None:
        database.suspend()
        self.register_action(f"Suspended MySQL database {database.id}")

    def _handle_resume(self, database) -> None:
        database.resume()
        self.register_action(f"Resumed MySQL database {database.id}")
```

plugins/modules/database_postgresql_v2.py
```diff
@@ -324,12 +324,46 @@ def _handle_absent(self) -> None:
...
     def exec_module(self, **kwargs: Any) -> Optional[dict]:
         """Entrypoint for token module"""
         state = kwargs.get("state")
 
         if state == "absent":
             self._handle_absent()
+        if state == "suspend":
+            self._handle_suspend()
+        if state == "resume":
+            self._handle_resume()
```
zliang-akamai, line 366:
```suggestion
        elif state == "suspend":
            self._handle_suspend()
        elif state == "resume":
            self._handle_resume()
```

ezilber-akamai (commented):
Overall the changes look good! I found a few things that were causing some issues with the integration tests. One thing is that `resume` and `suspend` are missing from the choices for the `state` field in the `databse_postgresql_v2` module.

ezilber-akamai (approved):
Nice work! Tests are passing locally.

zliang-akamai (approved):
LGTM!

PR #681, opened 2025-05-23
Firewall settings modules

zliang-akamai:
## 📝 Description

This is to add `firewall_settings` and `firewall_settings_info` module.

## ✔️ How to Test

You **might** need to enable default firewall for your account to test it.

**Note that once it's enabled for an account, it cannot be turned off, which means you will have to permanently have a default firewall for each of the resources in you account.**

```bash
curl --request PUT \
     --url https://api.linode.com/v4beta/networking/firewalls/settings \
     --header 'accept: application/json' \
     --header "authorization: Bearer $TOKEN" \   
     --header 'content-type: application/json' \
     --data '
{
    "default_firewall_ids": { 
        "public_interface": YOUR_DEFAULT_FIREWALL_ID,
        "vpc_interface": YOUR_DEFAULT_FIREWALL_ID,
        "linode": YOUR_DEFAULT_FIREWALL_ID,
        "nodebalancer": YOUR_DEFAULT_FIREWALL_ID
    }                 
}

'
```

Testing the modules:
 
```bash
make test-int TEST_SUITE=firewall_settings
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ Name | Description |
 [linode.cloud.domain_info](./docs/modules/domain_info.md)|Get info about a Linode Domain.|
 [linode.cloud.domain_record_info](./docs/modules/domain_record_info.md)|Get info about a Linode Domain Records.|
 [linode.cloud.firewall_info](./docs/modules/firewall_info.md)|Get info about a Linode Firewall.|
+[linode.cloud.firewall_settings_info](./docs/modules/firewall_settings_info.md)|Get info about a Linode Firewall Settings.|
 [linode.cloud.image_info](./docs/modules/image_info.md)|Get info about a Linode Image.|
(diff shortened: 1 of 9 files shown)


lgarber-akamai:
This looks great, nice work!

For the null `default_firewall_ids.X` issue you mentioned: it looks like the `_serialize(...)` method is excluding the keys of null fields from [FirewallSettingsDefaultFirewallIDs](https://github.com/linode/linode_api4-python/blob/proj/enhanced-interfaces/linode_api4/objects/networking.py#L220C7-L220C41), which is preventing `handle_updates(...)` from detecting a diff on those keys. Would it make sense to use [include_none_values, always_include, or put_class](https://github.com/linode/linode_api4-python/blob/032f29408c13d00e1b985f19f82eed7669ce7437/linode_api4/objects/serializable.py#L62-L80) in the Python SDK to address this?

PR #684, opened 2025-05-30
Update IP docs for Linode Interfaces

zliang-akamai:
## 📝 Description
Add an additional fields for Linode Interfaces.


zliang-akamai:
@ezilber-akamai The lint issue seems to be on another module not changed in this PR... Maybe we can fix it in a separate PR?

zliang-akamai:
The fix for the lint issue will go to dev branch in https://github.com/linode/ansible_linode/pull/685

PR #693, opened 2025-07-17
Added `account_settings` module

ezilber-akamai:
## 📝 Description

Added support for `GET /account/settings` and `PUT /account/settings` via `account_settings` module, and accompanying integration tests and documentation.

Shoutout to @jriddle-linode for the code!

## ✔️ How to Test

### Integration Tests
`make test-int TEST_SUITE="account_settings"`

### DIFF
--- a/plugins/modules/account_settings.py
+++ b/plugins/modules/account_settings.py
@@ -50,29 +50,11 @@ SPEC = {
             "The value must be a Longview subscription ID or null for Longview Free."
         ],
     ),
-    "managed": SpecField(
-        type=FieldType.bool,
-        description=[
-            "Our 24/7 incident response service. This robust, multi-homed "
-            "monitoring system distributes monitoring checks to ensure that "
-            "your servers remain online and available at all times. "
-            "Linode Managed can monitor any service or software stack "
-            "reachable over TCP or HTTP. Once you add a service to Linode Managed, "
-            "we'll monitor it for connectivity, response, and total request time."
-        ],
-    ),
     "network_helper": SpecField(
         type=FieldType.bool,
         description=[
             "Enables network helper across all users by default "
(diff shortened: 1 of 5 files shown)


jriddle-linode (approved):
LGTM

tests/integration/targets/account_settings/tasks/main.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+    - name: Get account_settings
+      linode.cloud.account_settings:
+        state: 'present'
+      register: info
+
+    - name: Assert account_settings response
+      assert:
+        that:
+          - info.account_settings.longview_subscription is not none
```
lgarber-akamai, line 14: I think `None` is a valid value for longview_subscription, so it might make sense to make this just check for the key

ezilber-akamai: Good catch! Fixed that and added the remaining fields to the assertion.

plugins/modules/account_settings.py
```diff
@@ -0,0 +1,148 @@
...
+            "for new Linodes and Linode Configs."
+        ],
+    ),
+    "object_storage": SpecField(
+        type=FieldType.string,
+        description=[
+            "A string describing the status of this account's "
+            "Object Storage service enrollment."
+        ],
+    ),
```
lgarber-akamai, line 77: `object_storage` and `managed` look to be read-only/managed through separate endpoints: https://techdocs.akamai.com/linode-api/reference/put-account-settings

ezilber-akamai: Those fields are excluded in the `MUTABLE_FIELDS` list. Do we want to exclude them from the spec entirely?

lgarber-akamai (approved):
Looks great!

PR #732, opened 2025-12-04
Fix handle_updates None filtering issue

lgarber-akamai:
## 📝 Description

This bug fixes an oversight in #731 that caused None values to not be filtered from the request body for `handle_update` operations without diff overrides.

## ✔️ How to Test

### Integration Testing

```
make test-int TEST_ARGS="-v database_postgresql_v2_engine_config"`
```


ezilber-akamai (approved):
Thanks for fixing this! Tests pass locally.

yec-akamai (approved):
Works well for me!

PR #739, opened 2025-12-22
Added support for Private Image Sharing

ezilber-akamai:
## 📝 Description

Added support for Private Image Sharing features.

> **_NOTES:_**
> * While `label` is an updatable field for Image Share Groups, it must serve as the unique identifier in the Ansible module and thus cannot be updatable.
> * The same is true for Image Share Group Tokens, as `valid_for_sharegroup_uuid` is not a unique field since a user can create multiple tokens for the same Image Share Group. Since `label` is the only updatable field for this resource in the API and it must serve as the unique identifier for the Ansible module, no fields are mutable in Ansible for this module.
> * The same is also true for Image Share Group Members, as `label` is the only updatable field but is also the identifying field.

## ✔️ How to Test

The following steps assume that you have pulled down this PR locally and ran `make install`. You must also run the following command: `export TEST_API_VERSION=v4beta`

Several tests for this feature require multiple accounts. In order for these tests to not be skipped, set `LINODE_PRODUCER_API_TOKEN` and `LINODE_CONSUMER_API_TOKEN` in your environment to two different accounts.

### Integration Tests
* `make test-int TEST_SUITE="image_share_group_basic"`
* `make test-int TEST_SUITE="image_share_group_info"`
* `make test-int TEST_SUITE="image_share_group_list"`
* `make test-int TEST_SUITE="image_share_group_token"`
* `make test-int TEST_SUITE="image_share_group_token_info"`
* `make test-int TEST_SUITE="image_share_group_token_list"`
* `make test-int TEST_SUITE="image_share_group_member"`
* `make test-int TEST_SUITE="image_share_group_member_info"`
* `make test-int TEST_SUITE="image_share_group_member_list"`
* `make test-int TEST_SUITE="image_share_group_image_list"`
* `make test-int TEST_SUITE="consumer_image_share_group_info"`
    * This test requires `LINODE_PRODUCER_API_TOKEN` and `LINODE_CONSUMER_API_TOKEN` to be set.
* `make test-int TEST_SUITE="consumer_image_share_group_image_list"`
    * This test requires `LINODE_PRODUCER_API_TOKEN` and `LINODE_CONSUMER_API_TOKEN` to be set.

> **_NOTE:_** For the `token` and `member` tests, try running them with `LINODE_PRODUCER_API_TOKEN` and `LINODE_CONSUMER_API_TOKEN` unset to verify that the tests get skipped properly.

### DIFF
--- a/plugins/modules/image_share_group_token.py
+++ b/plugins/modules/image_share_group_token.py
@@ -1,0 +1,161 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+"""This module contains all the functionality for Image Share Group Tokens for a Consumer."""
+
+from __future__ import absolute_import, division, print_function
+
+import ansible_collections.linode.cloud.plugins.module_utils.doc_fragments.image_share_group_token as docs
+from typing import Optional, Any
+from ansible_collections.linode.cloud.plugins.module_utils.linode_helper import (
+    filter_null_values,
+)
+from ansible_collections.linode.cloud.plugins.module_utils.linode_common import (
+    LinodeModuleBase,
+)
+
+from ansible_collections.linode.cloud.plugins.module_utils.linode_docs import (
+    global_authors,
+    global_requirements,
+)
+
+from ansible_specdoc.objects import (
+    FieldType,
+    SpecDocMeta,
+    SpecField,
+    SpecReturnValue,
+)
+from linode_api4 import ImageShareGroupToken
+
+SPEC = {
+    "label": SpecField(
+        type=FieldType.string,
+        description=["This Image Share Group Token's unique label."],
+        required=True,
+    ),
+    "valid_for_sharegroup_uuid": SpecField(
+        type=FieldType.string,
+        description=["The UUID of the Image Share Group that this token is valid for."],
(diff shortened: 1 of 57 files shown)


plugins/modules/image_share_group_token.py
```diff
@@ -0,0 +1,184 @@
...
+                msg="failed to create Image Share Group Token: {0}".format(
+                    exception
+                )
+            )
+
+    def _handle_present(self) -> None:
+        label = self.module.params.get("label")
+        token = self._get_image_share_group_token_by_label(label)
+
+        if not token:
```
lgarber-akamai: I wonder if it'd make sense to raise an error here if the token's `valid_for_sharegroup_uuid` diverges from the user-specified value, similar to what we do in the handle_updates function for non-mutable fields

ezilber-akamai: Great point! I added a check for that and added a test case for this as well. Thank you!

lgarber-akamai (approved):
Tests are all passing on my end, great work!

vshanthe (approved):
Tests Passing , Great work!

PR #755, opened 2026-02-09
domain_record TTLs are not rounded to closest valid values when diffing

mgwoj:
## 📝 Description

`ttl_sec` from `domain_record` is accepting any integer value, but API is rounding it to nearest accepted value, causing diff on update. As a solution we also perform this rounding locally.

## ✔️ How to Test

`ansible-test integration domain_record`


dawiddzhafarov (approved):

jriddle-linode (approved):
LGTM!

PR #757, opened 2026-02-13
Add support for increased block storage volume limits

dawiddzhafarov:
## 📝 Description

This PR adds support for configuring up to 64 devices (volumes/disks) per Linode Instance Configuration, depending on the Linode's RAM. 
Previously you could define 8 slots: `sda-sdh`. 
Now you can define `sda-sdz`, `sdaa-sdaz` and `sdba-sdbl`.

There was also a bug in the logic, when if you have skipped consecutive device from configuration, it was added anyway as a `Null` object.

## ✔️ How to Test

In order to utilize the logic for increased block storage volume limits, make sure to use `proj/block-storage-volume-limit-increase` branch inside `linode_api4-python` repository. (I have used: `pip install -e .`) 

To run unit tests (1 new unit test case):
`make test-unit`

To run Integration tests:
`make integration-test TEST_SUITE=instance_volumes_limit`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,4 @@
+# TODO: Update python-sdk dependency to latest version when merging proj/block-storage-volume-limit-increase branch.
 git+https://github.com/linode/linode_api4-python.git@proj/block-storage-volume-limit-increase#egg=linode-api4
 polling==0.3.2
 ansible-specdoc>=0.0.19
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-linode-api4>=5.39.0
+git+https://github.com/linode/linode_api4-python.git@proj/block-storage-volume-limit-increase#egg=linode-api4
 polling==0.3.2
 ansible-specdoc>=0.0.19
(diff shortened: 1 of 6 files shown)


yec-akamai (commented):
When running the integration test locally, it failed at `Create a Linode instance with 16 volumes`: `/linode_api4/objects/linode.py\", line 1307, in config_create\n    device_map[device_names[i]] = {\"volume_id\": d.id}\n               ~~~~~~~~~~~~^^^\n\nIndexError: list index out of range\n"`

yec-akamai (approved):
Tests all passed locally, nice work!

requirements.txt
```diff
@@ -1,3 +1,3 @@
-linode-api4>=5.39.0
+git+https://github.com/linode/linode_api4-python.git@proj/block-storage-volume-limit-increase#egg=linode-api4
```
yec-akamai, line 2: nit: add a TODO comment here as a reminder to revert it

ezilber-akamai (approved):

PR #763, opened 2026-03-02
Support Alerts Configuration in Instance Module

yec-akamai:
## 📝 Description

Support configuring both legacy and ACLP alerts in linode instance module. 

## ✔️ How to Test

You'll need to test against devcloud when running the integration test. 

```
make test-int TEST_SUITE=instance_alerts
```


plugins/modules/instance.py
```diff
@@ -387,6 +387,80 @@
...
+        type=FieldType.integer,
+        editable=True,
+        description=[
+            "The amount of disk IO operation per second required to trigger an alert. "
+            "If the average disk IO over two hours exceeds this value, "
+            "we'll send you an alert. "
+            "If set to 0, this alert is disabled. "
+            "Update only. "
+        ],
+    ),
```
zliang-akamai, line 443: Since we never supported legacy alerts before... do we want to only support the newer alerts and remove these legacy ones?

yec-akamai: The legacy ones work for different purpose and will not be removed or deprecated. Customer will use both legacy and aclp alerts as they need. I think we should add them

zliang-akamai (approved):
Looks good to me

plugins/modules/instance.py
```diff
@@ -951,6 +1033,23 @@ def _create_instance(self) -> dict:
...
+            )
+            if provided_disallowed:
+                self.fail(
+                    msg=(
+                        "Only alerts.system_alerts and alerts.user_alerts "
+                        "are supported during instance creation; "
+                        "remove: {0}"
+                    ).format(", ".join(provided_disallowed))
+                )
+
```
zliang-akamai, line 1052: Just a thought, do we really need to validate them? I think we might defer the responsibility of validation to the API...

zliang-akamai:
The test couldn't pass because I don't have system_alert in my account, but manual tests with empty array work well, and legacy alert update also works well.

plugins/modules/instance.py
yec-akamai: I think by validating it we can reduce the chance of sending bad request, and the error is more expectable

ezilber-akamai:
> The test couldn't pass because I don't have system_alert in my account, but manual tests with empty array work well, and legacy alert update also works well.

I also have the same issue. Is this something we can get added to our accounts or no?

yec-akamai:
> > The test couldn't pass because I don't have system_alert in my account, but manual tests with empty array work well, and legacy alert update also works well.
> 
> I also have the same issue. Is this something we can get added to our accounts or no?

Interesting, can both of you try the shared devcloud account if possible? Otherwise I can ask Lena to check the tags for your personal dev cloud account. we should have added it to our accounts though

ezilber-akamai:
> > > The test couldn't pass because I don't have system_alert in my account, but manual tests with empty array work well, and legacy alert update also works well.
> > 
> > 
> > I also have the same issue. Is this something we can get added to our accounts or no?
> 
> Interesting, can both of you try the shared devcloud account if possible? Otherwise I can ask Lena to check the tags for your personal dev cloud account. we should have added it to our accounts though

Same thing with that account. Maybe it is missing the customer tag?

ezilber-akamai (approved):

PR #769, opened 2026-03-23
TPT-4208: Implement support for NB Front-End IP & 40Gbps

dawiddzhafarov:
## 📝 Description

This PR adds support for Frontend IP in VPC for NodeBalancers and `type` attribute when creating or reading NodeBalancers.

## ✔️ How to Test

Run integration tests:

```
make TEST_SUITE="nodebalancer_basic" test-int 
```

```
make TEST_SUITE="nodebalancer_firewall" test-int
```

```
make TEST_SUITE="nodebalancer_frontend_vpc" test-int
```

```
make TEST_SUITE="nodebalancer_list" test-int
```

```
make TEST_SUITE="nodebalancer_node" test-int
```

```
make TEST_SUITE="nodebalancer_populated" test-int
```

```
make TEST_SUITE="nodebalancer_premium" test-int
```

```
make TEST_SUITE="nodebalancer_reordered_tags" test-int
```

```
make TEST_SUITE="nodebalancer_stats" test-int
```

```
make TEST_SUITE="nodebalancer_type_list" test-int
```

```
make TEST_SUITE="nodebalancer_backend_vpc" test-int
```

```
make TEST_SUITE="nodebalancer_vpc_list" test-int
```

```
make TEST_SUITE="nodebalancer_premium_40gb" test-int
```

```
make TEST_SUITE="nodebalancer_udp" test-int
```

### DIFF
--- a/plugins/modules/nodebalancer.py
+++ b/plugins/modules/nodebalancer.py
@@ -283,6 +283,11 @@ linode_nodebalancer_spec = {
         editable=True,
         description=["A list of configs to apply to the NodeBalancer."],
     ),
+    "type": SpecField(
+        type=FieldType.string,
+        description=["The type of this NodeBalancer."],
+        choices=["common", "premium", "premium_40gb"],
+    ),
 }
 
 
@@ -309,7 +314,7 @@ SPECDOC_META = SpecDocMeta(
             sample=docs.result_configs_samples,
         ),
         "nodes": SpecReturnValue(
-            description="A list of configs applied to the NodeBalancer.",
+            description="A list of nodes applied to the NodeBalancer.",
             docs_url="https://techdocs.akamai.com/linode-api/reference/get-node-balancer-node",
             type=FieldType.list,
             sample=docs.result_nodes_samples,
@@ -402,6 +407,7 @@ class LinodeNodeBalancer(LinodeModuleBase):
                 "label",
                 "firewall_id",
                 "tags",
+                "type",
             }
         }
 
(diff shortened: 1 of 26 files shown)


plugins/modules/nodebalancer.py
```diff
@@ -402,9 +490,25 @@ def _create_nodebalancer(self) -> Optional[NodeBalancer]:
...
+        vpcs_legacy = self.module.params.get("vpcs")
+
+        if backend_vpcs is not None:
+            params["backend_vpcs"] = backend_vpcs
+        elif vpcs_legacy is not None:
+            self.warn(
+                "The 'vpcs' field is deprecated and will be removed in a future major version. "
+                "Use 'backend_vpcs' instead."
+            )
+            params["backend_vpcs"] = vpcs_legacy
```
mawilk90, line 510: Should it not be `params["vpcs"]`? Or it is intentional to avoid error when `vpcs` field is removed completely in a future release?

dawiddzhafarov: Yep, that's intentional - when the field is removed and API does not support, we want to ensure we pass the param that is working and that it does not make our users to stop using this like they usually do (depending on when we decide to drop the `vpcs` attribute).
So it doesn't break suddenly in case they switch this param off.
Plus we know that the new param is a desired way of providing it, so it's less risk

mawilk90:
The tests work on Prod env with Lena's fixes ([here](https://github.com/linode/ansible_linode/pull/778)) merged locally -> approved

mawilk90 (approved):

ezilber-akamai (approved):

PR #772, opened 2026-03-26
Fix firewall update logic for `-1`

zliang-akamai:
## 📝 Description

`-1` represents explicit no firewall now. We need to handle it in the update logic.

## ✔️ How to Test
Install the collection in this branch
```bash
make install
```
Then run this playbook twice:
```bash
ansible-playbook /path/to/the/test_playbook.yaml
```

```yaml
- name: Test Firewall ID
  hosts: localhost
  collections:
    - linode.cloud
  tasks:
      linode.cloud.instance:
        label: example-linode
        type: g6-nanode-1
        region: us-central
        image: linode/arch
        root_pass: ReplaceWithAStrwediedekro3ouijqklongPassword123!
        state: present
        firewall_id: -1
        booted: false
      register: create_linode
  environment:
    LINODE_API_VERSION: v4beta

```

### DIFF
--- a/plugins/modules/instance.py
+++ b/plugins/modules/instance.py
@@ -1138,7 +1138,7 @@ class LinodeInstance(LinodeModuleBase):
         """
         firewall_id = self.module.params.get("firewall_id")
         # Nothing to do
-        if firewall_id is None:
+        if firewall_id in (None, -1):
             return
 
         # Resolve the expected firewall; fail if firewall doesn't exist
--- a/tests/integration/targets/instance_firewall_minus_one/tasks/main.yaml
+++ b/tests/integration/targets/instance_firewall_minus_one/tasks/main.yaml
@@ -1,0 +1,60 @@
+- name: instance_firewall_minus_one
+  block:
+    - set_fact:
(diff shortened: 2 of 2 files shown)


plugins/modules/instance.py
```diff
@@ -1138,7 +1138,7 @@ def _update_firewall(self) -> None:
         """
         firewall_id = self.module.params.get("firewall_id")
         # Nothing to do
-        if firewall_id is None:
+        if firewall_id in (None, -1):
```
mgwoj, line 1141: Should we have a test for this case?

mgwoj (approved):

plugins/modules/instance.py
lgarber-akamai: @zliang-akamai Will this case be covered by the existing tests once the change rolls out in the API?

lgarber-akamai (approved):

yec-akamai (approved):
```
make test-int TEST_SUITE=instance_firewall_minus_one
```
Test passed locally!

PR #778, opened 2026-04-09
TPT-4417: Resolve `nodebalancer` module ordering and sync issues; support in-place config updates

lgarber-akamai:
## 📝 Description

This pull request introduces retry and sync logic to account for the the propagation of NodeBalancer configs and nodes that occurs asynchronously on the backend.

This change also parallelizes the node instance creations in the `nodebalancer_populated` test to allow for quicker iteration.

## ✔️ How to Test

```
make test-int TEST_ARGS="-v $(ls tests/integration/targets/ | grep -i nodebalancer | tr '\n' ' ' )"
```

### DIFF
--- a/plugins/modules/nodebalancer.py
+++ b/plugins/modules/nodebalancer.py
@@ -585,9 +585,11 @@ class LinodeNodeBalancer(LinodeModuleBase):
     def _handle_configs(self) -> list:
         """Updates the configs defined in new_configs under this NodeBalancer"""
 
-        new_configs = self.module.params.get("configs") or []
-        if not new_configs:
+        configs_param = self.module.params.get("configs")
+        if configs_param is None:
             return []
+
+        new_configs = configs_param
 
         to_create, to_update, to_delete = self._plan_configs(new_configs)
 
@@ -670,7 +672,7 @@ class LinodeNodeBalancer(LinodeModuleBase):
 
     def _sync_config_nodes(
         self,
-        config: Optional[NodeBalancerConfig],
+        config: NodeBalancerConfig,
         nodes: Optional[List[dict]],
     ) -> None:
         """Sync nodes against the given config if provided."""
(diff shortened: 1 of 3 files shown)


plugins/modules/nodebalancer.py
```diff
@@ -420,7 +421,9 @@ def _create_config(
         Creates a config with the given kwargs within the given NodeBalancer.
 
         Retries config_create(...) on 400s while a conflicting config's
-        deletion is still propagating.
+        deletion is still propagating. If a matching remote surfaces while
+        the caller asked for a recreate, the remote is deleted and polling
+        continues until the create succeeds.
```
lgarber-akamai: The modules manages all of its configs if `configs` is specified, so this shouldn't cause any weird behavior that wasn't already occurring. We should consider supporting updates in the future though 🙂

plugins/modules/nodebalancer.py
```diff
@@ -524,57 +582,104 @@ def _check_config_exists(
...
+
+            successful_requests = (
+                successful_requests + 1
+                if len(self._node_balancer.configs) == num_configs
+                else 0
+            )
+
+            return successful_requests >= consecutive_threshold
+
+        poll_condition(_synced, step=2, timeout=180)
```
lgarber-akamai: For some reason the backend propagation sometimes also causes issues in the configs response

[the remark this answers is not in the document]
lgarber-akamai: Will be caught by a try-catch above.

lgarber-akamai:
Copilot lied about the async tests haha. Reverting that bit.

plugins/modules/nodebalancer.py
mawilk90: Sometimes `nodebalancer_populated` fails due to "failed to wait for condition: timeout period expired" on `Remove a config from the NodeBalancer` task - that's the reason I guess

zliang-akamai (approved):

yec-akamai (approved):
LGTM

PR #788, opened 2026-04-29
TPT-4284:ansible implement support for reserved ip for ipv4

mgwoj:
## 📝 Description

Implements support for Linode Reserved IPv4 addresses in `ansible_linode`.

New modules:

reserved_ip — create, update tags, and delete reserved IP addresses
reserved_ip_info — retrieve details for a single reserved IP by address
reserved_ip_list — list and filter all reserved IPs on an account

Updated modules:

instance — new ipv4 parameter to assign a reserved IP at creation time (create-only; excluded from update path to prevent handle_updates errors)
nodebalancer — new ipv4 parameter to assign a reserved IP at creation time (create-only; excluded from create params when None and popped before handle_updates)

Other changes:

Updated ip_info and ip doc fragments with new fields: reserved, tags, interface_id, assigned_entity, vpc_nat_1_1
Added integration tests: reserved_ip_basic, reserved_ip_info, reserved_ip_list, instance_reserved_ip, nodebalancer_reserved_ip
Pinned requirements.txt to linode_api4 @ git+https://github.com/linode/linode_api4-python.git@proj/reserved-ips

Generated/updated docs via make gendocs

## ✔️ How to Test

```
make test-unit

# Modified modules (regression check)
make TEST_SUITE="ip_info instance_basic nodebalancer_basic" test-int

# New reserved IP functionality
make TEST_SUITE="reserved_ip_basic reserved_ip_info reserved_ip_list instance_reserved_ip nodebalancer_reserved_ip" test-int
```


README.md
```diff
@@ -50,6 +50,7 @@ Name | Description |
 [linode.cloud.object_keys](./docs/modules/object_keys.md)|Manage Linode Object Storage Keys.|
 [linode.cloud.placement_group](./docs/modules/placement_group.md)|Manage a Linode Placement Group.|
 [linode.cloud.placement_group_assign](./docs/modules/placement_group_assign.md)|Manages a single assignment between a Linode and a Placement Group.|
+[linode.cloud.reserved_ip](./docs/modules/reserved_ip.md)|Manage a Linode Reserved IPv4 Address.|
```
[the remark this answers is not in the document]
mgwoj: Added missing doc files

zliang-akamai:
Some findings:
- We may need to update `ip` or `reserved_ip` module for the updated `/networking/ips` behavior to support promoting an IP to be a reserved IP. I haven't verified if it can work this way, but the gap does exist. For cleanest solution, we might need another brand new module for this. Optionally, we can also state that this feature is intentionally unsupported because users can still use generic API call module for it, your call.
- Should we declare the behavior of `linode.cloud.reserved_ip` that it is not idempotent when creating by region? Re-running the documented example will allocate another billable reserved IP each run.

These are mostly for discussion, the overall changes looks great!

zliang-akamai (approved):

tests/integration/targets/instance_reserved_ip/tasks/main.yaml
```diff
@@ -0,0 +1,59 @@
...
+        r: "{{ 1000000000 | random }}"
+
+    - name: Create an instance with the reserved IP
+      linode.cloud.instance:
+        api_token: "{{ api_token }}"
+        ua_prefix: "{{ ua_prefix }}"
+        label: "ansible-test-inst-{{ r }}"
+        region: us-ord
+        type: g6-nanode-1
+        image: linode/alpine3.19
```
mawilk90: This image version is deprecated, pls use e.g. linode/alpine3.23

mawilk90: Moreover, running `make TEST_SUITE="ip_info instance_basic nodebalancer_basic" test-int` (regression check) `ip_info` fails due to the same reason

docs/modules/ip.md
```diff
@@ -32,7 +42,36 @@ Allocates a new IPv4 Address on your Account. The Linode must be configured to s
 | `linode_id` | <center>`int`</center> | <center>Optional</center> | The ID of a Linode you have access to that this address will be allocated to.   |
 | `public` | <center>`bool`</center> | <center>Optional</center> | Whether to create a public or private IPv4 address.   |
 | `type` | <center>`str`</center> | <center>Optional</center> | The type of address you are requesting. Only IPv4 addresses may be allocated through this operation.  **(Choices: `ipv4`)** |
-| `address` | <center>`str`</center> | <center>Optional</center> | The IP address to delete.  **(Conflicts With: `linode_id`,`public`,`type`)** |
+| `address` | <center>`str`</center> | <center>Optional</center> | The IP address to update or delete. Required when updating an existing IP (e.g., promoting to reserved) or when deleting (state=absent).  **(Conflicts With: `linode_id`,`public`,`type`)** |
+| `reserved` | <center>`bool`</center> | <center>Optional</center> | Whether this IP address should be reserved. Setting to true promotes an existing allocated IP to a reserved IP via PUT /networking/ips/{address}. Requires the address parameter.   |
```
mawilk90: API Spec says that _"when 'reserved' is true, one of 'region' or 'linode_id' must be set"_ so shouldn't we also add a `region` as one of optional fields?

tests/integration/targets/ip_basic/tasks/main.yaml
mawilk90, line 99: Deletion of instance does not delete the reserved IP so please add a dedicated step/task to remove it

mawilk90:
What about `/tags` endpoints specified in the API Spec? Should we also cover them here?

tests/integration/targets/ip_basic/tasks/main.yaml
mgwoj: added

mawilk90:
There are some conflicts to be resolved but generally it looks good to me

mawilk90 (approved):

mawilk90:
@mgwoj, I've just noticed that PR goes directly to `dev`, could you please create a dedicated `project` branch?

plugins/modules/ip.py
```diff
@@ -92,10 +141,135 @@ def __init__(self) -> None:
             ],
```
lgarber-akamai, line 139: The required_together value might need to be updated to properly account for the new cases. Not sure if this will make a difference in practice though 🤔

mawilk90 (approved):

mawilk90:
From my perspective, it looks fine - re-approved

lgarber-akamai (approved):

ezilber-akamai (approved):

PR #798, opened 2026-05-13
TPT-4439: Added disk_encryption to LKE Cluster pools

ezilber-akamai:
## 📝 Description

Added support for `disk_encryption` field in LKE Cluster pools (creation only)

## ✔️ How to Test

`make test-int TEST_SUITE="lke_cluster_basic"`

### DIFF
--- a/plugins/modules/lke_cluster.py
+++ b/plugins/modules/lke_cluster.py
@@ -183,6 +183,15 @@ linode_lke_cluster_node_pool_spec = {
         description=["Firewall ID for the Node Pool."],
         required=False,
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        editable=False,
+        choices=["disabled", "enabled"],
+        description=[
+            "Local disk encryption setting for this LKE node pool.",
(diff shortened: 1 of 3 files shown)


mawilk90 (approved):

plugins/modules/lke_cluster.py
```diff
@@ -183,6 +183,15 @@
...
     ),
+    "disk_encryption": SpecField(
+        type=FieldType.string,
+        editable=False,
+        choices=["disabled", "enabled"],
+        description=[
+            "Local disk encryption setting for this LKE node pool.",
+            "This option can only be set during creation.",
+        ],
+    ),
```
lgarber-akamai, line 194: Not directly related to this change but it looks like we may need to explicitly pass disk_encryption here: https://github.com/ezilber-akamai/ansible_linode/blob/b01c29a1f5961e714bbed8cd93b004505b1da023/plugins/modules/lke_cluster.py#L737

lgarber-akamai (approved):
Looks great, nice work!

PR #820, opened 2026-06-24
TPT-4478: Added support for Configurable VPC IPv4 Prefixes

ezilber-akamai:
## 📝 Description

Added support for Configurable VPC IPv4 Prefixes features.

## ✔️ How to Test

`make test-int TEST_SUITE="vpc_ipv4"`
`make test-int TEST_SUITE="vpc_default_ranges_info"`

### DIFF
--- a/plugins/modules/vpc.py
+++ b/plugins/modules/vpc.py
@@ -140,8 +140,10 @@ class Module(LinodeModuleBase):
     ) -> Tuple[bool, Any]:
 
         def normalize(
-            values: Iterable[dict[str, Any]],
+            values: Optional[Iterable[dict[str, Any]]],
         ) -> Set[Tuple[Tuple[str, Any], ...]]:
+            if not values:
+                return set()
             return {tuple(sorted(v.items())) for v in values}
 
         return (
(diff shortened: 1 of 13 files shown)


mawilk90 (approved):

plugins/modules/vpc.py
```diff
@@ -118,6 +132,23 @@ def __init__(self) -> None:
...
+    @staticmethod
+    def _diff_ipv4(
+        _key: str,
+        old_value: Any,
+        new_value: Any,
+    ) -> Tuple[bool, Any]:
+
+        def normalize(
+            values: Iterable[dict[str, Any]],
+        ) -> Set[Tuple[Tuple[str, Any], ...]]:
```
lgarber-akamai, line 144: Any chance we could add protective null guards for old_value and new_value here? Not sure if it makes any difference in practice but it'd probably be good to cover our bases 👍

ezilber-akamai: Fixed!

lgarber-akamai (approved):
LGTM, great work!

PR #868, opened 2026-09-08
build(deps-dev): downgrade ansible requirement from >=14.3.1 to >=10.7.0 for python 3.10 support

zliang-akamai:


yec-akamai (approved):

lgarber-akamai (approved):
Looks great, thanks for fixing this!

